PR #4, opened 2021-03-12
Remove content submodule

benmccann:
This at least gets the site working on Linux again. I don't have a Windows machine to be able to test

We can put site-kit in this monorepo next to fully remove the need for submodules

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .DS_Store
 /node_modules
 /sites/*/content
-/content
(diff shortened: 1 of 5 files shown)


benmccann:
I updated this one so that no longer requires a symlink. Will it work for you on WIndows now?

benmccann:
Looks like Rich may have pushed his own version of https://github.com/sveltejs/sites/pull/5

PR #11, opened 2021-03-18
make clear that content should be edited within the contents of this repository

antony:


README.md
```diff
@@ -2,8 +2,12 @@
...
-## Running
+## Setup
 
 ```
 git submodule update --init --recursive
 ```
+
+You can now edit the content within each directory in [sites](sites) and see it live on your local machine.
+
+Commit as normal and submodule content will be added to the corresponding Svelte project repositories.
```
benmccann, line 13: maybe add something about pushing and opening a PR?

benmccann: i'm guessing you need to create a branch first? I saw something about detached head in the chat which I wouldn't know how to deal with

Rich-Harris: i think the branch stuff probably goes without saying. in any case this is all temporary

PR #14, opened 2021-03-22
prebuild docs

Rich-Harris:


Conduitry:
There are a number of issues with this. One is that it breaks local development if you _aren't_ prebuilding, since the endpoints are gone, and the prebuilt JSON files don't exist.

Another is that the `Object.keys(pkg.dependencies)` throws an error, because there are now no prod dependencies.

Locally, if I change that to `pkg.devDependencies`, and then I run `CI=true pnpm build` and `pnpm start`, I'm able to see the site locally. (And running `pnpm build` without `CI=true` will also result in a build without the built JSON files.) But, even if it works, and even if it gets the site deployed faster, I think the loss of local development is serious enough to not do this.

What's the precedence between files in `static` and endpoints? Could we have an endpoint for local (non-prebuilt) development, but write out the `docs.json` file for Netlify deployment? I'm not sure what that will help, though, if the issue is that Netlify is aggressively caching things. We should first see what happens if the site is redeployed as-is with their cache cleared.

Rich-Harris:
closing as @Conduitry found the whole thing rather distasteful

PR #21, opened 2021-04-13
Add an FAQ section to the docs

danawoodman:
image

Add nav bar link and a section styled similarly to that of Sapper incuding section links

### DIFF
--- a/sites/kit.svelte.dev/src/routes/faq/index.svelte
+++ b/sites/kit.svelte.dev/src/routes/faq/index.svelte
@@ -1,0 +1,137 @@
+<script context="module">
+	export async function load({ fetch }) {
+		const faqs = await fetch('faq.json').then(r => r.json());
+		return { props: { faqs } };
+	}
+</script>
+
+<script>
+	export let faqs;
+</script>
+
+<svelte:head>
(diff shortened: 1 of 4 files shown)


sites/kit.svelte.dev/src/routes/faq/index.svelte
```diff
@@ -0,0 +1,137 @@
...
+			</h2>
+			{@html faq.html}
+		</article>
+	{/each}
+	<p>
+		See also the <a href="https://svelte.dev/faq" rel="external">Svelte FAQ</a> for questions relating
+		to Svelte directly.
+	</p>
+</div>
+" /> -->
```
benmccann, line 36: this looks like an extra line that should be deleted? I don't see the start that would match up to it and it doesn't appear in the Sapper version: https://github.com/sveltejs/sapper/blob/master/site/src/routes/faq/index.svelte

benmccann, line 36:
```suggestion
```

danawoodman: Not sure why that was added, some auto-correct via Prettier I missed, sorry!

PR #24, opened 2021-04-17
chore: add additional inputs to deploy docs workflow

pngwn:
https://github.com/sveltejs/sites/runs/2371293908?check_suite_focus=true


lukeed:
holy blob

PR #17, opened 2021-04-13
#1 Home page - New style 

rafaelcamaram:
As suggested by https://github.com/sveltejs/kit/issues/244, I'm applying the new styles on the home page. This PR aims basically update the existing components with the new style and then, in the future, I'll create new PRs in order to add the missing sections.

Figma: https://www.figma.com/file/f87psJnCeqMbUPAQABmCrN/svelte?node-id=19%3A176

### Tests
- [X] Desktop
- [X] Mid-size
- [X] Tablet
- [X] Mobile

### Preview
desktop

mid-size

mobile

### Notes

- When I was changing some files, it was formatted in order to organize the order of `script`, `html` and `style` tags -- so there are some diffs only due to this reorganization
- Does any know what would be the correct links for the `On writing`, `On overhead` and `On reativity` buttons?

PR #29, opened 2021-04-23
Redesign

Rich-Harris:
Picking up where @rafaelcamaram left off. Closes #17


rafaelcamaram:
I just noticed that the buttons are slightly different and have also lost their arrow animation - is it expected or is it still in progress?

Screen Shot 2021-04-23 at 14 11 06
Screen Shot 2021-04-23 at 14 11 03

Rich-Harris:
It's expected — the buttons got huge in some situations, and it looked a bit odd

PR #32, opened 2021-04-23
Improve permalink a11y

geoffrich:
As instructed by @benmccann, moving https://github.com/sveltejs/site-kit/pull/35 to the new sites repo.

The original intention behind this PR was to improve a11y and fix some Axe violations on svelte.dev related to permalinks. To do this, I needed to update some of the site-kit components. I've done so and applied the a11y fixes to kit.svelte.dev.

Summary of the changes:
- Remove aria-hidden from permalinks. This fixes the Axe violation [ARIA hidden element must not contain focusable elements](https://dequeuniversity.com/rules/axe/4.0/aria-hidden-focus?application=AxeChrome)
- Add visually hidden text to each permalink so assistive tech knows what the link is
- Make focus visible when focusing permalinks and edit links. Previously there was either no focus outline (on edit links) or the element still had opacity 0 (permalinks)
- Move permalink to the end of the heading, so that the heading text is read first. This required tweaking some of the positioning CSS for the permalink

I was unable to fix all the permalinks, since the docs markdown is now being rendered via a [GH action](https://github.com/sveltejs/action-deploy-docs). If these changes look good, I'll make a corresponding PR there to update the permalink rendering.

I'd still like to fix the permalinks on the svelte.dev docs, but I think it's best to wait until it's moved into the site monorepo.


packages/site-kit/base.css
```diff
@@ -566,3 +566,16 @@ input[type='checkbox']::after {
 input[type='checkbox']:checked::after {
 	left: calc(100% - 1em + 2px);
 }
+
+/* visually hidden, but accessible to assistive tech */
+.visually-hidden {
```
benmccann, line 571: I'm actually not much of a Svelte expert despite working on it a fair bit. Is there a reason to put this here instead of in `Permalink.svelte`?

geoffrich: Yeah, I did this intentionally. `visually-hidden` is a common [utility class](https://piccalil.li/quick-tip/visually-hidden) used to hide content visually but still expose it to screen readers. It won't just be used in the Permalink component -- for instance, if I update the rendered markdown from the API to use that class, it will need to be available globally.

It could also be used in the future to hide other text, e.g. like hiding `<h1>`s on the [Svelte docs](https://github.com/sveltejs/svelte/blob/a39fc8de728ba292523828e1cf6f6254f125c640/site/src/routes/docs/index.svelte#L22).

PR #37, opened 2021-04-28
Fix offset anchor in FAQ

geoffrich:
This PR fixes the offset anchors in the FAQ so that the heading linked to is not covered by the header. 

For example, [this link](https://kit.svelte.dev/faq#hmr) looks like this in production (the "How do I use HMR with SvelteKit?" is covered by the header)...
image

And like this with my changes applied...
image

The implementation was copied from the Docs component.

I also fixed the permalink focus styles in the FAQ.

### DIFF
--- a/packages/site-kit/components/Docs.svelte
+++ b/packages/site-kit/components/Docs.svelte
@@ -246,7 +246,7 @@
 	.content :global(.offset-anchor) {
 		position: relative;
 		display: block;
-		top: calc(-1 * (var(--nav-h) + var(--top-offset) - 1rem));
+		top: calc(-1 * var(--top-offset));
 		width: 0;
 		height: 0;
 	}
--- a/sites/kit.svelte.dev/src/routes/faq/index.svelte
+++ b/sites/kit.svelte.dev/src/routes/faq/index.svelte
@@ -72,7 +72,7 @@
 	.faqs :global(.offset-anchor) {
 		position: relative;
 		display: block;
-		top: calc(-1 * (var(--nav-h) + var(--top-offset) - 1rem));
+		top: calc(-1 * var(--top-offset));
 		width: 0;
 		height: 0;
 	}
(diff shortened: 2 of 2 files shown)


benmccann:
What browser are you on?

I'm on Chrome. For me the header disappears, so when I click [the link](https://kit.svelte.dev/faq#hmr) you provided I don't see an issue

geoffrich:
I'm on Firefox -- checked Chrome, and you're right that the header disappears there. However, iOS Safari shows the same issue as Firefox (see below)
hmr

FWIW, this is the same offset method the svelte.dev docs and FAQ use.

benmccann:
How does your PR look on Chrome? Will it leave a gap in the nav disappears?

geoffrich:
With my PR the gap will always be there, whether or not the header is visible. You can try it out on the [deploy preview](https://deploy-preview-37--kit-svelte-dev.netlify.app/faq#hmr). Weirdly enough, I'm seeing some inconsistent Chrome behavior -- occasionally it still shows the header on page load when navigating to a fragment.

This should behave the same way as the Kit docs, where the gap is always there.

We could possibly put the open state of the nav in a store and adjust the gap accordingly, though that might be over-complicating things. I'm also not sure if that would work consistently, or if it would be too late by the time the page is loaded.

benmccann:
The deploy preview site is working better for me. In Chrome it looks just like [your screenshot](https://user-images.githubusercontent.com/4992896/116433141-11f91380-a7fe-11eb-9a14-ed95f92b8de5.png)

Do you think we can get rid of the bit of extra gap? I.e. make it so that the text "SvelteKit is currently in beta..." is not showing and it starts directly with the linked heading

geoffrich:
How does [this](https://deploy-preview-37--kit-svelte-dev.netlify.app/faq#hmr) look? It's not an exact science; I got it as close as I can without cutting the heading off on mobile. I also made the changes on the Kit docs for consistency.

Desktop:
image

Mobile:
IMG_3746

benmccann:
thanks!

PR #38, opened 2021-04-29
Add h1s and fix heading order

geoffrich:
Several pages on the SvelteKit site did not have an h1, resulting in an [axe a11y violation](https://dequeuniversity.com/rules/axe/4.1/page-has-heading-one). This PR adds those headings and makes them only visible to screen readers so the pages look the same.

Also, the use of an h3 in the Hero component resulted in an [invalid heading order](https://dequeuniversity.com/rules/axe/4.1/heading-order) violation. There didn't seem to be much benefit to this being a heading, so I changed it to a div and updated the styles so it looked the same.

### DIFF
--- a/packages/site-kit/components/Hero.svelte
+++ b/packages/site-kit/components/Hero.svelte
@@ -12,7 +12,7 @@
 	<div class="hero-container">
 		<div class="hero-text">
 			<img alt="{title} logotype" width="400" height="50" class="logotype" src={logotype} />
-			<h3>{tagline}</h3>
+			<div class="tagline">{tagline}</div>
 		</div>
 
 		<div class="hero-image">
@@ -55,7 +55,7 @@ linear-gradient(0deg, #DBE7EF, #DBE7EF);
 		mix-blend-mode: multiply;
 	}
 
-	.hero-banner h3 {
+	.hero-banner .tagline {
 		margin-top: 0;
 		position: relative;
 		font-size: 2rem;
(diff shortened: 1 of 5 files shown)


packages/site-kit/components/Hero.svelte
```diff
@@ -12,7 +12,7 @@
 	<div class="hero-container">
 		<div class="hero-text">
 			<img alt="{title} logotype" width="400" height="50" class="logotype" src={logotype} />
```
benmccann, line 14: should the `alt` here just be `{title}` instead of `{title} logotype`?

sites/kit.svelte.dev/src/routes/docs/index.svelte
```diff
@@ -22,4 +22,5 @@
 	<meta name="Description" content="Complete documentation for SvelteKit">
 </svelte:head>
 
+<h1 class="visually-hidden">Docs</h1>
```
benmccann, line 25: Maybe "Documentation"? We probably just went with "Docs" in the nav to save space

benmccann:
The hero change makes sense to me. For the heading change, I noticed that the FAQ page has a visible `h1`. I wonder if we can make it more consistent by making the `h1` visible on the Migrating and Docs pages as well

packages/site-kit/components/Hero.svelte
geoffrich: Updated in latest commit.

sites/kit.svelte.dev/src/routes/docs/index.svelte
geoffrich: Updated in latest commit.

geoffrich:
@benmccann here's how the pages look with the visible h1s.

image

image

Issue #33, opened 2021-04-24
Touch events on nav menu navigating to home


bedewilson:
On mobile devices - and when not on the kit.svelte.dev home page - touching the menu to open it navigates to the home page.

The Nav component has the following code in it:
```
<ul
    class:open
    on:touchstart|passive|capture={intercept_touchstart}
    on:mouseenter={() => (open = true)}
    on:mouseleave={() => (open = false)}
>
```
I think the `passive` event modifier clashes with the `.preventDefault()` call in the `intercept_touchstart` function. This may be breaking the intercept functionality.

I experienced this on Android 10/Chrome 90 and the responsive design mode in Firefox 88.

PR #40, opened 2021-04-30
Intercept touch events for responsive menu only

### LINKED ISSUE #39: Navbar links require double click on landscape touch devices
Currently when in the desktop layout (i.e. not the dropdown menu) on touch devices, links have to be touched twice to navigate.

The `<Nav>` component has the following code to intercept touch events and open the responsive menu.
```javascript
function intercept_touchstart(event) {
  if (!open) {
    event.preventDefault();
    event.stopPropagation();
    open = true;
  }
}
```
The problem is that this function runs on both mobile and desktop layouts. So for touch devices showing the desktop layout, the first touch toggles the `open` variable (which has no visible effect), and the second touch is passed through to the link which then navigates.

bedewilson:
Designed to fix #39. Currently you have to double click Navbar links when in the desktop layout on touch devices. This is because touch events are intercepted to toggle the responsive menu - even in the desktop layout when it is never shown. The first touch event toggles the menu, and the second touch event is passed to the link. This problem affects phones in landscape, and tablets.

To address this I am determining the current layout by checking the window's `innerWidth`, and using this to prevent touch events being intercepted in the desktop layout.

_However_ this does feel like a bit of a kludge as the mobile/desktop breakpoint now needs to be set in both the script and style sections. I would be interested in your thoughts on this.

**To Do**
Having the home link in the menu overlap the drop down button is what necessitates these work arounds. Addressing this in the design would simplify the component and also fix the bug, and allow some of the issues in the Lighthouse audit to be fixed also.

- [ ] Simplify design to remove overlapping links
- [ ] Remove `<div>` under `<ul>` element to fix a11y issues for screen readers
- [ ] Re-add passive event modifier on scroll listener to improve performance once `preventDefault` is not required


packages/site-kit/components/Nav.svelte
```diff
@@ -20,8 +20,14 @@
...
+	let window_inner_width;
+	function is_mobile() {
+		if (window_inner_width < 768) return true;
+		return false;
+	}
+
 	function intercept_touchstart(event) {
-		if (!open) {
+		if (!open && is_mobile()) {
 			event.preventDefault();
```
benmccann, line 31: I wonder if we even need this line and the next. Even after the menu is open or when I'm on my laptop, clicking the link for the current page does nothing

bedewilson: I see what you mean - I think this is another issue. The menu is closed when the page store changes (i.e. on navigation). I'm presuming that SvelteKit doesn't change this store if the navigation is to the same page. So this would leave the menu open when clicking the current page's link.

```javascript
	// hide nav whenever we navigate
	page.subscribe(() => {
		open = false;
	});
```

benmccann: It's not really an issue though, is it? The behavior seems fine. So I'm wondering if we can take advantage of this behavior to simplify the code and just remove the `preventDefault` and `stopPropagation`

bedewilson: Ok sorry misunderstood. The `preventDefault` is needed to stop the menu navigating to the home page when you touch to open it. This is what was happening in #33. But I can't see a need for `stopPropagation` so that can probably be removed. I will test that now.

benmccann: I still don't understand why `preventDefault` is needed. Let's say you're on `/faq`. Then that's what the menu will say. If you click it then SvelteKit's router will handle the click, see the URL matches the current URL, and do nothing. So it seems to me like you could just let the click event be handled. Am I missing something?

bedewilson: I can make a repro later today.

If you're on `/faq` and you touch the dropdown link the menu opens. Because the menu opens on touch though, it opens before the click event fires. Regardless of the current page, it's always the `Home` button that is directly under the cursor when the menu is open. The click event then fires and it navigates to `Home`.

image

image

bedewilson: Here is the current state of this PR but with the `preventDefault` removed - http://kit-no-preventdefault.surge.sh/. I have put my proposed approach in the _Todo_ section of the PR, but open to any alternative suggestions.

Many thanks for your time reviewing this.

packages/site-kit/components/Nav.svelte
```diff
@@ -20,10 +20,15 @@
 		open = false;
 	});
 
+	let window_inner_width;
+	function is_mobile() {
```
benmccann, line 24: I think it'd be better to call this `is_small_screen` or something. Because it could be triggered on desktop as well

bedewilson: Sounds good. My only other thought was to shift the menu down so the links don't overlap (like the mockup below) - that would remove the need for all this code and the `preventDefault`, and allow us to use the passive event modifier on the scroll listener that had previously caused an issue in #33 (which was added to address an issue in the Lighthouse audit).

If you'd prefer I can do that instead - otherwise I'll just make the change to `is_small_screen` and wrap it up.

kit-mockup

packages/site-kit/components/Nav.svelte
```diff
@@ -20,10 +20,15 @@
...
+		if (window_inner_width < 768) return true;
```
Mlocik97, line 25:
```js
if () {
    return true;
}
return false;
```

is nonsense,... you can just simply write:

`return (window_inner_width < 768)`

Mlocik97, line 25:
```js
if () {
    return true;
}
return false;
```

is nonsense,... you can just simply write:

`return (window_inner_width < 768)`

Rich-Harris:
Thank you — this issue is fixed on the current site, which has a slightly different design for the nav, so I'll close this

PR #45, opened 2021-05-15
Upgrade SvelteKit to v1.0.0-next.107

nico-bachner:
## Changes
There have been two notable changes to SvelteKit since v1.0.0-next.87:

1. Vite is no longer a peer dependency of kit but a regular dependency. Therefore, I removed it from the dependencies.
2. `$layout.svelte` and `$error.svelte` are now to follow the naming convention `__layout.svelte` and `__error.svelte`

Also, the Svelte config is now supposed to be an ES Module, no longer a CommonJS file.

## Test
I ran the site in dev mode to assure nothing was broken, but all seems to be fine. Local build also worked without issue.

I assume Netlify will deploy a build of this PR, so that should double-check if the build is broken.


nico-bachner:
And Netlify is a go.

sites/kit.svelte.dev/.gitignore
```diff
@@ -2,5 +2,6 @@
 node_modules
 .env
 /.svelte
```
benmccann, line 4:
```suggestion
```

PR #49, opened 2021-07-31
Add Svelte link to Kit website

Mlocik97:


benmccann:
this seems like it might be out of place since it's text and the others are images

Mlocik97:
I was thinking that on right side, there are "external" sources, resp. anything else than `kit.svelte.dev`, should I move it to middle?

benmccann:
that doesn't seem right either. my inclination would be to make it a svelte logo

Mlocik97:
I got that idea about using logo too. but the problem is Svelte and SvelteKit have same logo. For user it can be confusing?

Mlocik97:
How about this?

benmccann:
Can you share a screenshot?

Mlocik97:
Snímka obrazovky z 2021-08-01 15-39-26
Snímka obrazovky z 2021-08-01 15-39-56

benmccann (approved):

Mlocik97:
will this be merged? Btw. after rebase, this PR is totally mess...

benmccann:
Fixed version of your PR here: https://github.com/sveltejs/sites/pull/62

I think you might just need to delete your `sites` repo and start over

PR #50, opened 2021-08-14
Preview docs

pngwn:
This points the site at a version of the API hosted on `localhost:3456` when `pnpm dev:docs` is run allowing a local preview of kit documentation changes.

See: https://github.com/sveltejs/kit/pull/2202

I don't know how Vite's default env handling works, there is probably a better way of doing this.


netlify.toml
```diff
@@ -4,4 +4,5 @@
   functions = ".netlify/functions/"
 
 [build.environment]
-  NPM_FLAGS="--prefix=/dev/null"
\ No newline at end of file
+  NPM_FLAGS="--prefix=/dev/null"
+  NODE_VERSION="12.22"
```
benmccann, line 8: either is fine, but I might use a newer version like `14.16.0`

benmccann (approved):
lgtm. thanks for this!

netlify.toml
pngwn: Yeah, lets try that.

PR #53, opened 2021-08-28
Run pre and post scripts

benmccann:
I saw the `prebuild` and `postbuild` steps:

https://github.com/sveltejs/sites/blob/0f195fbc127b3c209842e30d476c034d09238a18/sites/kit.svelte.dev/package.json#L8

It took me a long time to figure out why they weren't running on my machine. In some version of `pnpm` the pre and post scripts stopped being enabled by default I guess. Enable with https://pnpm.io/next/cli/run#enable-pre-post-scripts

Are these `prebuild` and `postbuild` steps left over or are they actually being used? I'm not sure why they're configured to run only on the CI and I'm not sure why we'd need to be checking out the Kit repo if we're getting the content from the API


Rich-Harris:
Pretty sure we can just delete those `prebuild` and `postbuild` scripts? Looks like they date back to a time before the docs were delivered via the API

Rich-Harris:
Ah, just read the whole comment. Yep, we agree. Will just remove the scripts

Rich-Harris:
huh. netlify did NOT like that

PR #59, opened 2021-10-09
fix body not using correct font

### LINKED ISSUE #35: body on kit-site is missing font 'Overpass'
Just noticed this while reading the kit-docs.
body-css

So the body on kit-site looks like:
sveltekit body

While on the svelte-site we have the 'Overpass' as body-font:
svelte body

saitheninja:
Fixes #35

### DIFF
--- a/packages/site-kit/base.css
+++ b/packages/site-kit/base.css
@@ -59,8 +59,7 @@
 }
 
 body {
-	--font: 'Overpass', sans-serif;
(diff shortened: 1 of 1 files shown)


benmccann:
thanks! I checked this out locally and agree it looks better

nimmolo:
I'm not seeing this fix on the live site yet. 
`body` is still using `var(--font-body)` instead of `var(--font)`

benmccann:
I think Netlify deploys are failing until https://github.com/netlify/cli/pull/3533 is merged. Hopefully they will do that soon

benmccann:
I think this should be deployed now

PR #61, opened 2021-10-16
Update URLs

stranger26:
If user trying to navigate from kit.svelte.dev/*/* by clicking on the `NavItem` it will be redirected to kit.svelte.dev/*/NavItem instead of kit.svelte.dev/NavItem.


benmccann:
I believe this is working just fine today due to the `base` tag and there aren't any broken links on the site. This seems like a good change anyway though, so I will go ahead and merge it

PR #64, opened 2021-11-16
Generate hero properly sized with avif support

benmccann:
This is probably one of the things we can do to most speed up the site

The current hero image is 1440x1440, but only needs to be 960x960:

https://github.com/sveltejs/sites/blob/4cb5b1368f5097d020357dbbbf05234742efd640/sites/kit.svelte.dev/src/routes/index.svelte#L28

Serving avif in browsers that support it will result in a smaller file size as well


Rich-Harris:
Where did the 960x960 come from? The image can be as large as 800x800 given the current layout — is there a danger that 960x960 could look pixellated on retina screens?

benmccann:
The 960x960 is defined in the current code. I'm not sure how it was arrived at

Where do you get the 800x800 number from?

I actually only see it show up as 480x600 when I visit the site:

Screenshot from 2021-11-16 11-36-16

benmccann:
Hmm. I think there's a Chrome bug. If I drag the window smaller, the image stays the same size, but Chrome tells me it's 800 × 600 px

Rich-Harris:
Sorry, I meant 800x600. There are a couple of different breakpoints
 
image

benmccann:
Ok, I pushed a new version with the corrected dimensions

Rich-Harris:
Well now it's _definitely_ too small! When I questioned 960x960, it was because 800x600 logical pixels means 1600x1200 _actual_ pixels on a screen with a device pixel ratio of 2 (i.e. retina laptop screens). Generally you can get away with it being a bit less than 2 depending on the type of image (hence the 1440x1440) but 1 seems like it would result in pixellation or blurriness, no?

PR #65, opened 2021-11-17
various changes to site-kit

Rich-Harris:
so far:

* [x] move nav styles into the `<Nav>` component, so less stuff needs to be copied into each project
* [x] add some common icons, so they don't need to be duplicated across projects
* [ ] remove the `font-size: 62.5%`, since it messes with your own CSS (e.g. it breaks svelte-knobby)


packages/site-kit/components/Nav.svelte
```diff
@@ -77,14 +77,16 @@
 			<li class="hide-if-desktop" class:active={$current === ''}>
 				<a sveltekit:prefetch href="/">{home}</a>
 			</li>
-			<slot name="nav-center" />
+			<div class="nav-center">
```
benmccann, line 80: not sure if this fixes it, but I noticed that Lighthouse gives an accessibility warning for the nav, so might be worth taking a look at while this section is being updated

Rich-Harris: because there's a `<div>` as a child of a `<ul>`? hmm. this won't fix it, no. though it won't make it worse either. should probably be addressed in a follow-up — going to merge this and release because the `npm link` workflow is killing me (seriously, npm, why do you nuke all linked packages as soon as you install something new? so dumb) but will open an issue for that

Rich-Harris:
ended up not changing the `font-size` because it will have far-reaching effects — instead i updated svelte-knobby to use `px` everywhere

benmccann:
Just FYI, it turns out pnpm overrides works way better than npm link for a lot of stuff

PR #66, opened 2021-11-18
make docs components more content-agnostic

Rich-Harris:
The `<Docs>` component currently makes certain assumptions about the structure of docs content that limit how it can be used.

In particular, I'm working on `svelte-cubed` docs, and want to include components in those docs. This means that rather than pulling static HTML from an API, it needs to be able to display docs that come from `.svelte.md` files etc. This PR is designed to make that possible


packages/site-kit/components/docs/Main.svelte
```diff
@@ -171,45 +75,8 @@
...
-			padding: var(--top-offset) 0 6.4rem 3.2rem;
-			font-family: var(--font);
-			overflow-y: auto;
-			height: 100%;
-			bottom: auto;
-			width: 100%;
-		}
-
+	@media (min-width: 832px) {
+		/* can't use vars in @media :( */
```
benmccann, line 79: I think you can with `env`? But not my area of expertise

https://developer.mozilla.org/en-US/docs/Web/CSS/env()

Rich-Harris: 🤯

pngwn: Could use a [postcss plugin](https://github.com/postcss/postcss-custom-media) for this, it follows the spec.

packages/site-kit/components/docs/Contents.svelte
```diff
@@ -0,0 +1,257 @@
...
+								class="nested subsection"
+								class:active={subsection.path === path}
+								href={subsection.path}
+							>
+								{subsection.title}
+							</a>
+						{/each}
+					{/each}
+				</li>
+			{/each}
```
pngwn, line 71: Should we make these nested ul > li > ul for better semantics and probably some minor a11y improvements?

packages/site-kit/components/docs/Main.svelte
```diff
@@ -225,18 +93,18 @@
 		}
 	}
 
-	.content h2 {
+	.content :global(h2) {
 		margin-top: 8rem;
 		padding: 2rem 1.6rem 4rem 0.2rem;
 		border-top: 2px solid #ddd;
 		/* color: var(--text); */
```
pngwn, line 100: Remove the comments if they're not being used.

packages/site-kit/components/docs/Main.svelte
```diff
@@ -370,57 +237,57 @@
...
 	}
 
-	section :global(blockquote) {
+	.content :global(section) :global(blockquote) {
 		/* color: #ff3e00; */
-		color: rgba(0,0,0,0.7);
+		color: rgba(0, 0, 0, 0.7);
 		background-color: rgba(255, 62, 0, 0.1);
 		/* background-color: #f6f6f6; */
 		/* border-left: 4px solid var(--flash); */
```
pngwn, line 265: More comments

packages/site-kit/components/docs/Main.svelte
```diff
@@ -370,57 +237,57 @@
...
-	section :global(blockquote) :global(:last-child) {
+	.content :global(section) :global(blockquote) :global(:last-child) {
 		margin-bottom: 0;
 	}
 
-	section :global(blockquote) :global(code) {
+	.content :global(section) :global(blockquote) :global(code) {
 		/* background: hsl(204, 100%, 95%) !important; */
 		background: #d5e2ea;
 		/* color: hsl(204, 100%, 50%); */
```
pngwn, line 281: comments

packages/site-kit/components/docs/Main.svelte
```diff
@@ -370,57 +237,57 @@
...
 
-	section :global(a):hover {
+	.content :global(section) :global(a):hover {
 		text-decoration: underline;
 	}
 
-	section :global(a) :global(code) {
+	.content :global(section) :global(a) :global(code) {
 		color: inherit;
 		/* background: none !important; */
```
pngwn, line 290: comment

packages/site-kit/components/docs/Contents.svelte
Rich-Harris: probably, yeah — there might have been some reason why it wasn't done that way in the first place but will try it

packages/site-kit/components/docs/Main.svelte
Rich-Harris: Personally I'm inclined towards the magic numbers just because it's one less moving part. That comment has been there for a long time, and we've managed

packages/site-kit/components/docs/Contents.svelte
pngwn: Original it was a flat list of nav elements so was trickier. When I redid the docs formatting I changed it to nested so this would be easier. We just never got round to making use of it.

pngwn (approved):
lgtm!

PR #68, opened 2021-11-18
make h1 bigger

Rich-Harris:
right now the typographical hierarchy is all wrong

## before

image

## after

image


benmccann (approved):
seems fine to me. though I didn't have any objection to the before version either :smile:

PR #71, opened 2021-11-21
Add note about node version requirement

gtm-nayan:
The `Unexpected token '.'` error has appeared quite a few times on discord and recently on the Kit issue tracker as well. Older node.js doesn't support optional chaining so starting the dev server fails with no changes to the starter template whatsoever. It isn't a great experience for new people if they have to go hunting for the fix.

I posted a screenshot on discord, that one had an extra line in it, but that should be fixed now, and it actually looks decent without the huge gap.

### DIFF
--- a/sites/kit.svelte.dev/src/routes/index.svelte
+++ b/sites/kit.svelte.dev/src/routes/index.svelte
@@ -80,6 +80,7 @@
 
 	<div slot="how">
 		<pre><code>
+<span class="comment"># You'll need Node.js v14 or higher</span>
 npm init <span class="orange-highlight">svelte@next</span> my-app
 cd my-app
 npm install
@@ -99,5 +100,10 @@ npm run dev -- --open
 
 	.orange-highlight {
 		color: var(--prime);
+	}
+
+	.comment {
+		color: var(--second);
+		font-style: italic;
 	}
 </style>


benmccann:
We specified Node 14 in the `engines` field of SvelteKit's `package.json`. That's supposed to stop people from running with old versions. If that's not working we should figure out why

Conduitry:
I think as of a while ago npm by default won't refuse to install when the `engines` requirement isn't met, but will instead only print a warning. This message, however, will likely get lost among a bunch of other innocuous warnings (e.g., about `fsevents` on non-macOS platforms, and about packages for prebuilt `esbuild` binaries for other platforms you're not currently on). npm (rightfully, I think) now leaves it up to the user whether they want to abort on unmet `engines` fields, rather than letting the package author decide for themselves.

So, in particular, I _do_ get a message from npm on Node 12 currently, but I also get a bunch of other messages. I don't know what the correct answer is here. If we do want to explicitly document this somewhere, I think it should be somewhere more visible than a C-style comment in a block of commands to be executed in the terminal.

sites/kit.svelte.dev/src/routes/index.svelte
```diff
@@ -80,6 +80,7 @@
 
 	<div slot="how">
 		<pre><code>
+// You'll need Node.js v14 or higher
```
benmccann, line 83: This isn't JS so should use bash-style commenting
```suggestion
# You'll need Node.js v14 or higher
```

gtm-nayan: Aah right, fixed now.

gtm-nayan:
I also opted for a more muted color and italic text so it looks more like a comment. Chrome says the contrast ratio (5.28) should be fine.
image

gtm-nayan:
> I think it should be somewhere more visible than a C-style comment in a block of commands to be executed in the terminal.

Wouldn't it be the most visible here since this is where most people will learn the installation process for the first time?

benmccann:
I don't get any warning using npm 6 and Node 12 on Ubuntu. If I create an `.npmrc` with `engine-strict=true` then it will fail. We try to do that today:

https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/skeleton/.npmrc
https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/default/.npmrc

But if I run `npm init svelte@next my-test-app` I don't see an `.npmrc` file being created in the new app, so I think there's some bug in `svelte-create` there

Issue #90, opened 2021-05-28
Site is not accessible with keyboard


ZerdoX-x:
**Describe the bug**
Site navigation broken due to [svelte-repl component](https://github.com/sveltejs/svelte-repl)

**To Reproduce**
1. Throw your mouse away.
2. Go to [svelte.dev](https://svelte.dev/).
3. Try to reach any interactive element such a navigation link.

**Expected behavior**
I could reach every (or at least header navigation) interactive element

**Severity**
Whole site keyboard navigation broken. You can use mouse only in this case.

**Additional context**
[sveltejs/svelte-repl/issues/163](https://github.com/sveltejs/svelte-repl/issues/163)

If you still want to support this feature (spaces appear by pressing tab):
1. Work on component. Improve it.
2. Some site workaround. For example how google does:
(go to https://www.google.com/search?q=test and press tab once)
![image](https://user-images.githubusercontent.com/49815452/119984129-76b0b500-bfda-11eb-9d1c-2a1792c1c86b.png)

Rich-Harris:
fixed via https://github.com/sveltejs/sites/pull/122, thanks @geoffrich!

PR #122, opened 2021-12-15
implement a11y fixes from svelte-repl

Rich-Harris:
adds the fixes from https://github.com/sveltejs/svelte-repl/pull/168, plus a couple of adjacent tidy-ups


geoffrich:
This should resolve https://github.com/sveltejs/sites/issues/90 and https://github.com/sveltejs/svelte-repl/issues/163

PR #124, opened 2021-12-15
only expose <Repl> from repl package

Rich-Harris:
not a big deal, but makes things a bit tidier


Rich-Harris:
(related: `lib/index.svelte` should probably be the `"."` export rather than the `"./index.svelte"` export... will try and remember to open an issue on Kit)

PR #125, opened 2021-12-15
Improve keyboard a11y for mobile nav

geoffrich:
This PR improves the keyboard and screen reader navigation experience in the mobile nav.

- Move button before the list of links. This way, when the nav is opened with the keyboard, the links are the next item in the focus order. Previously, they were before the button in the DOM and could be missed by screen reader and keyboard users.
-  Add `aria-expanded` attribute to convey the state of the button to screen readers when focused.
- Automatically close the nav when focus moves outside of it. Previously, focus would move further down the page, but the nav is still open so focus could be lost.

### DIFF
--- a/packages/site-kit/components/Nav.svelte
+++ b/packages/site-kit/components/Nav.svelte
@@ -8,6 +8,7 @@
 
 	let open = false;
 	let visible = true;
+	let nav;
 
 	// hide nav whenever we navigate
 	page.subscribe(() => {
@@ -30,15 +31,25 @@
 		last_scroll = scroll;
 		hash_changed = false;
 	}
+
+	function handle_focus() {
+		if (open && !nav.contains(document.activeElement)) {
+			open = false;
+		}
+	}
 </script>
 
-<svelte:window on:hashchange={handle_hashchange} on:scroll={handle_scroll} />
+<svelte:window
+	on:hashchange={handle_hashchange}
+	on:scroll={handle_scroll}
+	on:focusin={handle_focus}
+/>
 
 {#if open}
 	<div class="modal-background hide-if-desktop" on:click={() => (open = false)} />
 {/if}
 
-<nav class:visible={visible || open} class:open>
+<nav class:visible={visible || open} class:open bind:this={nav}>
 	<a
 		sveltekit:prefetch
 		href="/"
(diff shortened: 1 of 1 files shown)


benmccann:
Just FYI, because I know you are interested in this area, we've added a label for site accessibility issues: https://github.com/sveltejs/sites/issues?q=is%3Aopen+is%3Aissue+label%3Aaccessibility

Rich-Harris:
Thanks so much! I notice that (as is the case on the current site) if I open the menu with the keyboard then hit tab, it doesn't focus the links inside the newly opened menu, but instead moves focus outside the nav altogether (which closes it). Is that intentional?

geoffrich:
Hm, I don't think I'm seeing the same thing. The desired behavior is for the next item in the keyboard focus order to be the first menu link once you open the menu.

In Chrome/Firefox on Windows when I open the menu with the keyboard and hit tab, my focus is moved to the first link in the open menu and the menu stays open. Once I tab outside of the menu, it automatically closes. GIF below of using Chrome on this PR's [deploy preview](https://svelte-dev-git-fork-geoffrich-nav-a11y-svelte.vercel.app/):

keyboard-nav

What browser are you seeing this on?

Rich-Harris:
Ah, interesting — it works in Chrome, but not in Firefox. Haven't investigated why yet

geoffrich:
Even weirder -- it works in Firefox 95 on Windows for me. 

keyboard-nav-ff

[This is the code](https://github.com/geoffrich/svelte-sites/blob/96356bd6baa96d805d42599edbe63c97dab17850/packages/site-kit/components/Nav.svelte#L35-L38) that closes the modal when focus moves outside -- would you be able to tell me what `document.activeElement` is? If it's closing the nav on tab, then it must be moving focus to something outside the nav.

Conduitry:
As another data point: For me on Firefox 95 on Windows with the current site, pressing tab brings me to other items in the site in the background behind the nav area - but on the deploy preview for this branch, it brings me through the items in the nav area.

Rich-Harris:
What the actual hell? Fails for me in FF 94 and 95 on Monterey, both on the deploy preview and locally. Deeply, deeply weird. But I guess we worry about that another time

PR #130, opened 2021-12-15
one debouncing approach

Rich-Harris:
https://github.com/sveltejs/svelte-repl/pull/157

I actually think we can improve the experience by applying the debouncing to the error overlay and the importer specifically, rather than the whole thing. It's nice to get immediate feedback when typing


Rich-Harris:
closing in favour of #131

PR #131, opened 2021-12-16
debouncing

Rich-Harris:
WIP. This debounces messages (so you're not constantly distracted by syntax errors while you're mid-word), and the plan is to also debounce imports (so if you type `import x from 'some-package-name'` you don't end up fetching `so` and `some-` and `some-pac` etc from unpkg.

The goal is the same as #130, but bundling still happens immediately, which is nicer UX.

### DIFF
--- a/packages/repl/src/lib/Input/ModuleEditor.svelte
+++ b/packages/repl/src/lib/Input/ModuleEditor.svelte
@@ -16,6 +16,23 @@
 	export function focus() {
 		editor.focus();
 	}
+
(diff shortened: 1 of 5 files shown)


Rich-Harris:
Implemented the second half by adding an artificial delay to uncached imports. If you start typing

```js
import * as THREE from 'thr'
```

then by the time 250ms have elapsed, you'll have added the `ee`, and the initial request will be aborted.

PR #185, opened 2021-12-16
Improve color contrast of syntax highlighting

geoffrich:
The color contrast of the syntax highlighting in the code samples did not meet WCAG AA standards (4.5 for small text), which could make it hard to read for people with low vision. This PR updates the colors used to match that threshold. This reduces the count of [color contrast Axe violations](https://dequeuniversity.com/rules/axe/4.0/color-contrast) to 233 from 1493 instances.

Contrast changes (checked using https://contrast-ratio.com with #f6fafd background):
- code-comment: was 3.93, now 5.23
- code-function: was 4.01, now 4.5
- code-string: was 3.49, now 4.65
- code-number: was 2.84, now 4.89
- code-template-string: was 3.05, now 4.85

Before:
image

After: 
image

I made a similar PR against the old site repo (https://github.com/sveltejs/svelte/pull/5681), but it looks like those changes got lost at some point in the migration. This PR uses those changes as a base but leaves all colors as HSL and does not change tokens that now meet the 4.5 threshold.

### DIFF
--- a/packages/site-kit/code.css
+++ b/packages/site-kit/code.css
@@ -3,11 +3,12 @@
 :root {
 	--code-bg: var(--back-light);
 	--code-base: hsl(45, 7%, 35%);
-	--code-comment: hsl(210, 25%, 50%);
+	--code-comment: hsl(0, 0%, 41%);
 	--code-keyword: hsl(204, 88%, 35%);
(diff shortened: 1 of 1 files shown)


Rich-Harris:
thank you!

Issue #100, opened 2019-12-03
REPL sharing title and image


rlaferla:
1.  The shared image for the REPL is too big.  This can be annoying on chat channels like Discord.
2.  The actual title of the REPL should be used when sharing instead of "Cybernetically enhanced web apps".

![repl](https://user-images.githubusercontent.com/1135694/70056704-e55daf80-15a9-11ea-8088-91ab34ee2c04.png)

Rich-Harris:
closed via #223, thanks

PR #187, opened 2021-12-16
feat: show repl title on twitter meta

tanhauhau:
#100 for the twitter title. not sure about the images

### DIFF
--- a/sites/svelte.dev/src/lib/db/gist.js
+++ b/sites/svelte.dev/src/lib/db/gist.js
@@ -49,12 +49,26 @@ export async function create(user, gist) {
 
 /**
  * @param {string} id
- * @returns {Gist}
+ * @returns {Promise<Gist>}
  */
 export async function read(id) {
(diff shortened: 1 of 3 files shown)


Rich-Harris:
I wonder if at this point we should just fetch the full gist in `load`, rather than fetching metadata early then later fetching the content?

tanhauhau:
I wonder if we able to / should we differentiate boy user agent vs normal user to fetch the full gist?

Not sure how wasteful is fetching the content if it is only for generating thumbnail

Rich-Harris:
They get cached pretty aggressively I think, so it's probably not the thing to optimise for. The only real question is whether fetching the full gist is likely to slow navigation to the point that it feels sticky, which I think is generally unlikely. And it would allow us to show a preview of the code (in an inert textarea) while the REPL loads (in a follow-up PR).

Rich-Harris:
closing in favour of #222 and #223

PR #188, opened 2021-12-16
update acorn worker parser version to ecma 12

### LINKED ISSUE #141: Update ECMAScript version from 9 to 12 to sync with svelte?
I just noticed that acorn in `svelte-repl` runs with a setting of `ecmaVersion: 9` (2018) where svelte is using `ecmaVersion: 12` (2021). I'm not entirely sure if this is a bug or intended, but to me it seems like it should be in sync?

https://github.com/sveltejs/svelte-repl/blob/2d0ecdf288c117f93a208efaaf04f2e38240fb86/src/workers/bundler/plugins/commonjs.js#L17

tanhauhau:
fixes #141

### DIFF
--- a/packages/repl/src/lib/workers/bundler/plugins/commonjs.js
+++ b/packages/repl/src/lib/workers/bundler/plugins/commonjs.js
@@ -14,7 +14,7 @@ export default {
 
 		try {
 			const ast = acorn.parse(code, {
-				ecmaVersion: 12
+				ecmaVersion: 'latest'
 			});
 
 			const requires = [];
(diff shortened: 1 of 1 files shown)


packages/repl/src/lib/workers/bundler/plugins/commonjs.js
```diff
@@ -14,7 +14,7 @@ export default {
 
 		try {
 			const ast = acorn.parse(code, {
-				ecmaVersion: 9
+				ecmaVersion: 12
```
Rich-Harris, line 17: should this be 'latest', so we don't have to keep updating this?

Rich-Harris, line 17:
```suggestion
				ecmaVersion: 'latest'
```

PR #190, opened 2021-12-16
Start to cleanup setup instructions

benmccann:
Remove the old `docker-compose.yml`. If we want to run Supabase on Docker, the instructions for doing so are quite different and won't use this file: https://supabase.com/docs/guides/hosting/docker

Change `npm` to `pnpm`

Remove old instructions about linking `@sveltejs/site-kit` and `@sveltejs/svelte-repl` which is no longer necessary


Rich-Harris:
At some point we're going to want to set up a `todo` table to power `api.svelte.dev/todos` (https://github.com/sveltejs/api.svelte.dev/issues/8) which is why I felt it made more sense to keep it at the top level — it's not specific to just svelte.dev

benmccann:
Ok, I moved it back for now. That makes me wonder if the API should live in this monorepo as well

I also wonder if it's a good idea to have two different sites sharing the same database. If we don't want them each to have their own database, we could try to make the API the single source of truth for info coming from the database and do everything through it

Rich-Harris:
Not opposed to moving https://github.com/sveltejs/api.svelte.dev in here, but would want to know what @pngwn and @lukeed think.

I don't think there's much advantage in putting it between svelte.dev and Supabase though. It's just an extra hop (and another set of keys to look after — would need to make sure that only svelte.dev was able to make certain requests of the API) with no obvious benefit.

I can imagine a situation in which we added a feature to kit.svelte.dev that required you to log in, and at that point we might want to share the `user` table between two sites rather than maintaining duplicate tables. So even if we spun up a new database for `/todos` (which feels like overkill? not sure), we still wouldn't necessarily have a 1:1 site:db mapping

PR #198, opened 2021-12-17
Add ability to delete single gists

oskarhane:
For a some time I've wanted to be able to cleanup my gists in the apps list.
This PR adds this ability, by adding a delete button to each gist in the "Your apps" list.

I made the change as small and simple as possible, using the browsers built in `confirm` dialog.

**UPDATE 2021-12-19**:

I've pushed some changes to make this a more robust and proper solution:  
When a gist is deleted, the new gists list for the current page is re-fetched and loaded. If there are no gists for the current offset after a deletion, and we have a prev offset (i.e. last gist on last page deleted), we'll navigate to the prev page.

I also added a "Previous" pagination link to the bottom of the list.

A page state "deleteState" is introduced to make sure we don't open up for deletion during deletion on slower connections.

**/UPDATE**

### Some screenshots

On hover:  
Screen Shot 2021-12-17 at 16 27 56

Confirm:  
Screen Shot 2021-12-19 at 16 43 02

Pagination:  
Screen Shot 2021-12-19 at 16 42 40

### DIFF
--- a/db/setup.sql
+++ b/db/setup.sql
@@ -82,7 +82,7 @@ as $$
 		return query
 		select gist.id, gist.name, gist.created_at, gist.updated_at
 		from gist
-		where gist.userid = list_userid
+		where gist.userid = list_userid and gist.deleted_at is null
 		order by coalesce(gist.updated_at, gist.created_at) desc
 		limit list_count + 1
 		offset list_start;
@@ -129,7 +129,7 @@ as $$
 	begin
 		update gist
 			set name = gist_name, files = gist_files, updated_at = now()
-			where id = gist_id and userid = gist_userid
+			where id = gist_id and userid = gist_userid and deleted_at is null
 			returning id, name, files, userid into ret;
 
 		return ret;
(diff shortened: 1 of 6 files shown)


Rich-Harris:
Thanks — FYI I opened #202, we don't want to end up deleting data irretrievably. As well as adding a `deleted_at` column, it changes the signature of `gist.destroy` (there's no need to send all the gist metadata, just the `id`). Added a note in there that we will probably want the ability to bulk-delete, though that can come later.

It would probably be more RESTful if we were making a `DELETE` request against `/repl/[id].json` rather than `/app.json`, though admittedly I haven't thought about what that would look like in the bulk-delete case.

oskarhane:
Thanks @Rich-Harris, that makes sense. I still haven't gotten too familiar with the codebase yet.

I've updated this PR to build on your #202 and addressed the comments you had.

wdyt about the `window.location.reload()` after a successful deletion? 
I figured it was the simplest solution for now, but could read in the new list of gists for the current page number instead.

Rich-Harris:
Added the ability to bulk-delete apps in #214, so I'll close this — thank you

PR #200, opened 2021-12-17
fix tutorial nav arrows position

sharu725:
The tutorial arrows look fine when served locally but appears different in production.

image

This line should align them to the center.

image


Rich-Harris:
thank you!

PR #202, opened 2021-12-17
update deleted_at column rather than actually deleting

Rich-Harris:
we decided against actually deleting gists, when we do eventually get round to exposing the ability via the UI. will need to update the prod db before merging this

* add `deleted_at timestamptz` column to `gist`
* update functions


Rich-Harris:
merging this but it's incomplete — need to add a bulk delete function, will do that in a follow-up

PR #208, opened 2021-12-20
fix checkbox style

tanhauhau:
The checkbox style is removed in https://github.com/sveltejs/sites/pull/69/files

added back:

Screenshot 2021-12-20 at 7 49 08 PM

Screenshot 2021-12-20 at 7 49 30 PM

### DIFF
--- a/packages/site-kit/components/Checkbox.svelte
+++ b/packages/site-kit/components/Checkbox.svelte
@@ -1,0 +1,59 @@
+<script>
+	export let checked;
+</script>
+
+<input type="checkbox" bind:checked />
+
+<style>
+	input[type='checkbox'] {
+		/* display: block; */
+		position: relative;
+		height: 1em;
+		width: calc(100% - 0.6em);
+		max-width: 2em;
+		top: -2px;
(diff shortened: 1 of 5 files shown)


gtm-nayan:
Perhaps it could be renamed to ToggleSwitch or something similar to be clearer

packages/site-kit/components/Checkbox.svelte
```diff
@@ -0,0 +1,59 @@
+<script>
+	export let checked;
+</script>
+
+<input type="checkbox" bind:checked />
+
+<style>
+	input[type='checkbox'] {
```
gtm-nayan, line 8: The attribute selector shouldn't be needed since it's the only input inside the component?

Rich-Harris:
Thanks — is this checkbox style used anywhere else on the site? Feels like it should probably be part of the `repl` package rather than `site-kit`

Rich-Harris:
I'll close this in favour of #220 (which is a branch off this branch) — realised it makes sense to move `<InputOutputToggle>` inside the REPL. Thanks

PR #210, opened 2021-12-20
searchable gist

tanhauhau:
Allow gists to be searchable


sites/svelte.dev/src/routes/apps/index.svelte
```diff
@@ -55,6 +79,8 @@
 			</div>
 		</header>
 
+		<input placeholder="Search apps..." bind:value={searchValue} />
```
geoffrich, line 82: This needs a label for a11y -- either an actual `<label>` or an `aria-label` attribute.

Rich-Harris:
Merged #214 which implements search in a different way (traditional search input, no filter-as-you-type), so I'll close this

PR #214, opened 2021-12-20
apps page redesign

Rich-Harris:
Slightly more visually appealing, and a better use of space. In time we could augment these blocks with screenshots, if we build out the infrastructure for creating them. The highlighted block is being hovered

image


Rich-Harris:
Current status:

image

Searching works as soon as you hit 'enter'. Selecting apps...

image

...switches to a different mode, in which the links are no longer links. Clicking the 'delete' button will open a confirmation dialog, then delete the gists in question. Right now it reloads the page after doing so, because I just found a bug with `invalidate` which I'll raise on the Kit repo.

PR #215, opened 2021-12-20
Fix #212 Edit this page link in Tutorials not linked to the correct page.

### LINKED ISSUE #212: Edit this chapter links on svelte tutorial are broken
Looks like the variable isn't being set anymore

https://github.com/sveltejs/sites/blob/257ac05aea48bfb29ba61cc8205643de973e5e72/sites/svelte.dev/src/routes/tutorial/%5Bslug%5D/index.svelte#L76-L80

https://github.com/sveltejs/sites/blob/257ac05aea48bfb29ba61cc8205643de973e5e72/sites/svelte.dev/src/routes/tutorial/%5Bslug%5D/index.svelte#L157-L159

sharu725:
Right now `slug` doesn't give us the exact file name information with which we can create a link to the exact chapter in a section.

But until then users can be redirected to the tutorials index page - ttps://github.com/sveltejs/svelte/tree/master/site/content/tutorial

### DIFF
--- a/sites/svelte.dev/src/routes/tutorial/[slug]/index.svelte
+++ b/sites/svelte.dev/src/routes/tutorial/[slug]/index.svelte
@@ -74,7 +74,7 @@
 		'https://github.com/sveltejs/svelte/tree/master/site/content/tutorial';
 
 	$: selected = lookup.get(slug);
-	$: improve_link = ``;
+	$: improve_link = '';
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you, but if we're going to fix this we should _actually_ fix it — working on it over at https://github.com/sveltejs/action-deploy-docs/issues/20

PR #220, opened 2021-12-20
Fix checkbox

Rich-Harris:
Supersedes #208 — moves the input/output toggle into the `repl` package, which results in a more logical relationship between components and makes it easier to use `<Repl>`


Rich-Harris:
Marking as draft because there's some remaining details to iron out around `orientation`

PR #224, opened 2021-12-21
Icons

Rich-Harris:
Started updating some of the icons used on the site to match @vedam's Figma document

Before:

image

After:

image


benmccann:
these icons are still used on the new site. i'll leave this open for now for reference

PR #232, opened 2021-12-21
[fix] force sharp@^0.29.0

### LINKED ISSUE #230: `sharp` dependency cannot be installed on WSL
Installing dependencies in this repo currently fails on WSL, with an error during `sharp`'s postinstall script about not being able to create the directory `C:\Users\[username]\AppData\Roaming/npm-cache`. The solution to this seems to be to upgrade the version of `sharp` we're using. (I can reproduce this same error by installing `sharp@0.28.3` in an empty directory, but it does not happen for `sharp@0.29.3`, the current latest version.)

I don't know what would be involved with this, because this is currently hidden away in `vite-imagetools` -> `imagetools-core` -> `sharp`. We could try doing an override or using a pnpmfile, as it seems the only breaking change in 0.29 was dropping support for old Node versions.

Conduitry:
This resolves #230 for me - I'm now able to install dependencies. We'll see whether this breaks the preview deployment. If other folks aren't comfortable with this change, that's understandable, but it would certainly hamper my ability to contribute to this repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
 	"devDependencies": {
 		"@changesets/cli": "^2.14.1"
 	},
+	"pnpm": {
+		"overrides": {
+			"sharp": "^0.29.0"
+		}
+	},
 	"scripts": {
 		"release": "pnpm publish --tag=next --filter=\"@sveltejs/*\""
 	}
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,5 +1,8 @@
 lockfileVersion: 5.3
 
+overrides:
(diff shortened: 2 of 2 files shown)


benmccann:
I don't love this, but at least it unblocks you, so I guess I'd say go for it. I wouldn't mind if we could actually put comments in the file so we know why it was added and could remove it later. I filed an issue for it in imagetools: https://github.com/JonasKruckenberg/imagetools/issues/294. It looks like they have some test fixtures that would need to be updated to get the tests passing with an upgrade

benmccann (approved):

PR #233, opened 2021-12-21
Fix link to apps in REPL menu

oskarhane:
The link to the users REPL apps is wrong in the REPL menu.


oskarhane:
Addressed in https://github.com/sveltejs/sites/pull/234

PR #238, opened 2021-12-22
RTL Support (Logical Properties)

### LINKED ISSUE #228: Consider using logical CSS properties.
This change might be advantageous if you plan to support RTL languages with #207. Logical properties allow you to write dimensions for both LTR and RTL writing modes agonistically for the price of getting rid of the `top/left/bottom/right` terminology. At the time of writing this, they are supported in all major browsers. https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties

Tropix126:
Closes #228 adding roughly-tested RTL support.

### DIFF
--- a/packages/repl/src/lib/Checkbox.svelte
+++ b/packages/repl/src/lib/Checkbox.svelte
@@ -8,28 +8,29 @@
 	input[type='checkbox'] {
 		/* display: block; */
 		position: relative;
-		height: 1em;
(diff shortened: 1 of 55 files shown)


Rich-Harris:
Wonderful, thank you! Will be a challenge not to slip back into bad habits, but having this foundation will make it easier

PR #239, opened 2021-12-22
Revert logical floats

Tropix126:
Apologies on this; floats were the last thing I converted, so I didn't catch this in time. Appears that they are relatively new and not well supported yet, breaking the REPL on some browsers.


Rich-Harris:
That seems to have fixed it! Thank you

PR #245, opened 2021-12-23
Add skip link to Svelte and SvelteKit sites

geoffrich:
This PR adds a [skip link](https://webaim.org/techniques/skipnav/) to svelte.dev and kit.svelte.dev. This allows users navigating by keyboard or with a screen reader to bypass the main navigation, which they would otherwise need to tab through on every page load. It is hidden until it receives focus with the keyboard, at which point it becomes visible.

Points of note:
- ~I had to set `rel="external"` on the skip link. Not doing so made it not work the first time and reset focus to the top of the page, not the main content. Not sure if this is expected or a bug in SvelteKit~ This was a bug, and is no longer needed since latest SvelteKit fixes it
- The skip link needed a z-index set so that it appeared over the main nav.

Screenshot of how it appears when focused:

Screenshot of how the skip link appears when focused


benmccann:
Hmm, you shouldn't need `rel=external`. The behavior you described does sound like a bug

bluwy:
There's also coincidentally a [discussion](https://github.com/sveltejs/kit/discussions/3098) earlier about the bug as well that might be related.

geoffrich:
This is ready for review now. I was able to remove `rel="external"` from the skip link after the SvelteKit issue was fixed.

bluwy (approved):
Works fine for me

benmccann (approved):

PR #248, opened 2021-12-29
show ast output in repl

bluwy:
Updated mainly the `@sveltejs/repl` package. The `<Repl/>` component has a `showAst` prop to show the new ast output. Currently turned on for `/repl` only, other pages like tutorials and examples probably doesn't make sense.

I think the ast tree styling and a11y could use some improvement, but for now it's fairly usable.

Credits: https://astexplorer.net for design inspiration


Conduitry:
This appears to have broken the alternate input/output toggle on mobile. I'm getting a display wider than my screen that I need to scroll horizontally on.

bluwy:
> This appears to have broken the alternate input/output toggle on mobile. I'm getting a display wider than my screen that I need to scroll horizontally on.

That should be fixed now. The auto `scrollIntoView()` call was interfering with the mobile view. Found the neat transform trick for mobile toggle as well.

packages/repl/src/lib/index.svelte
```diff
@@ -175,6 +182,9 @@
 
 		register_module_editor(editor) {
 			module_editor = editor;
+			module_editor.cursorIndex.subscribe((index) => {
+				cursor_index.set(index);
+			});
```
bluwy, line 187: I'm subscribing it here without unsubscribe since there isn't another nice way to transfer the store state to `cursor_index`. It should be safe too since codemirror also has stuff that aren't unsubscribed.

packages/repl/src/lib/workers/compiler/index.js
```diff
@@ -31,13 +31,14 @@ const common_options = {
...
-		const { js, css } = svelte.compile(source, Object.assign({}, common_options, options));
+		const { js, css, ast } = svelte.compile(source, Object.assign({}, common_options, options));
 
 		return {
 			id,
 			result: {
 				js: js.code,
-				css: css.code || `/* Add a <sty` + `le> tag to see compiled CSS */`
+				css: css.code || `/* Add a <sty` + `le> tag to see compiled CSS */`,
+				ast
```
bluwy, line 41: Would transferring the AST from the worker, even with `showAst` false, affect the performance? Not super familiar with web workers, but if this is a concern, we could conditionally return the `ast`

packages/repl/src/routes/index.svelte
```diff
@@ -1,8 +1,8 @@
 <script>
 	import Repl from '@sveltejs/repl';
 	import { onMount } from 'svelte';
-	import '../../../site-kit/base.css';
-	import '../../../site-kit/code.css';
+	import '../../../site-kit/src/lib/base.css';
+	import '../../../site-kit/src/lib/code.css';
```
bluwy, line 5: The styles got moved previously, updated here.

dummdidumm:
Love this! I use the AST explorer site extensivly during my work on language tools but newer features are not present there because the Svelte version is too old. This would be solved with this feature and I could even look at different versions easily.

ignatiusmb:
Nicely done! I think it makes sense to enable it in `/examples` as well, people looking around for some Svelte code would likely be interested to see the AST output as well, since that's probably the next part newcomers would look through after `/tutorial`.

benmccann:
I've never used the AST and wouldn't know what to do with it. It's a pretty advanced feature. It sounds very cool though!! But I agree putting it only on REPL and leaving it off the examples since it's not for beginners

dummdidumm:
From the maintainers meeting: The AST is not public API, so we don't make it too visible and we want to make it clear that it's private API. So we decided to add a banner that this is private API.

bluwy:
Added a message at the bottom of the AST tab:
image
Wanted to keep it short and sweet, but let me know if we should expand/rephrase it.

dummdidumm:
Looks good to me, I'm not a native speaker so I'm not entirely sure if "may break at any point in time" is 100% correct, because the AST doesn't break, it's a change that's breaking - but don't know if that makes a difference in practice.

benmccann:
It didn't jump out at me as being wrong and I knew what it meant though is suppose updating "break" to " change" or otherwise rewording as you suggested would be slightly more accurate

packages/repl/src/lib/Output/AstView.svelte
```diff
@@ -0,0 +1,98 @@
...
+			{#if typeof ast === 'object'}
+				<ul>
+					<AstNode value={ast} {path_nodes} {autoscroll} collapsed={false} />
+				</ul>
+			{:else}
+				<p>No AST available</p>
+			{/if}
+		</code>
+	</pre>
+	<Message kind="info">The AST is not public API and may break at any point in time</Message>
```
bluwy, line 64:
```suggestion
	<Message kind="info">The AST is not public API and may change at any point in time</Message>
```

bluwy:
Updated "break" to "change". I think it sounds nicer and clear too.

dummdidumm (approved):

benmccann (approved):

Issue #73, opened 2021-11-30
h3 anchors disappear beneath the nav


Rich-Harris:
If you go to https://kit.svelte.dev/docs#routing-endpoints, the heading appears in the correct place:

![image](https://user-images.githubusercontent.com/1162160/144142505-1e66f9f4-607f-4643-8531-dd2245064cf2.png)

If you go to its child section, https://kit.svelte.dev/docs#routing-endpoints-body-parsing, the heading is hidden beneath the nav:

![image](https://user-images.githubusercontent.com/1162160/144142538-1e30f391-ccac-488b-b614-6cf6c2c5c31b.png)

PR #255, opened 2022-01-02
fix: h4s in docs overlay preceding text

### LINKED ISSUE #246: .content :global(h4) is overlaying previous paragraph
As mentioned in the title,  [padding-block-start: 10rem;  used in .content :global(h4) selector](https://github.com/sveltejs/sites/blob/5bf13ac400a4e01806ffd9eabd4ba047e111b51c/packages/site-kit/src/lib/components/docs/Main.svelte#L207) is overlaying partial of the previous paragraph. More details can be seen in the below screenshots.
Screen Shot 2021-12-23 at 10 18 33 PM

This is causing couple usability issues:
1. Last sentence of the previous paragraph cannot be selected with mouse. If there's a link in the last sentence, that cannot be clicked. 
2. Link of the current header shows up in the wrong location:
Screen Shot 2021-12-23 at 10 18 59 PM

Also note that currently when a link is opened, it's scrolled a bit lower from the top of the page, hence it's visible even when the top navigation bar is visible. If the `padding-block-start` is completely removed, then the opened link's title will be at the top of the page hidden behind the top navigation bar.

gtm-nayan:
Closes #246

Avoids #73 in a slightly different way, see https://nicolasgallagher.com/jump-links-and-viewport-positioning/demo/#method-B

### DIFF
--- a/packages/site-kit/src/lib/docs/Main.svelte
+++ b/packages/site-kit/src/lib/docs/Main.svelte
@@ -219,10 +219,6 @@
 		opacity: 0.7;
 	}
 
-	.content :global(h4 > .anchor) {
-		inset-block-start: 0.05em;
-	}
-
 	.content :global(h5) {
 		font-size: 2.4rem;
 		margin-block: 2em 0.5em;
--- a/packages/site-kit/src/lib/docs/Main.svelte
+++ b/packages/site-kit/src/lib/docs/Main.svelte
@@ -17,8 +17,8 @@
 				a.className = 'anchor';
 				a.href = `${$page.path}#${heading.id}`;
 				const span = document.createElement('span');
-				span.className = 'visually-hidden';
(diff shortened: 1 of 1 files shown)


packages/site-kit/src/lib/components/docs/Main.svelte
```diff
@@ -200,20 +200,23 @@
...
+	.content :global(h4::before) {
+		display: block;
+		content: ' ';
+		block-size: var(--nav-h);
+		margin-block-start: calc(-1 * var(--nav-h));
 	}
 
-	.content :global(h4 > .anchor) {
-		inset-block-start: 0.05em;
+	.content :global(h4 > em) {
```
gtm-nayan, line 218: Github seems to be messing up the diff here, more readable diff is
image

bluwy (approved):

PR #256, opened 2022-01-02
add emmet plugin to the repl

omar2205:
Adds [emmetio/codemirror-plugin ](https://github.com/emmetio/codemirror-plugin) to the repl editor, to add Emmet functionality.

### DIFF
--- a/packages/repl/package.json
+++ b/packages/repl/package.json
@@ -23,6 +23,7 @@
 	},
 	"type": "module",
 	"dependencies": {
+		"@emmetio/codemirror-plugin": "^1.2.4",
 		"@sveltejs/site-kit": "workspace:*",
 		"acorn": "^8.6.0",
 		"codemirror": "5",
--- a/packages/repl/src/lib/CodeMirror.svelte
+++ b/packages/repl/src/lib/CodeMirror.svelte
@@ -131,7 +131,7 @@
         CodeMirror = mod.default;
         emmet = (await import('@emmetio/codemirror-plugin')).default;
 			}
-      emmet(CodeMirror);
+      await emmet(CodeMirror);
       await createEditor(mode || 'svelte');
 			if (editor) editor.setValue(code || '');
 		})();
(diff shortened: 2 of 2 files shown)


omar2205:
hey @benmccann, can you sort this PR? reject or accept is fine.

benmccann:
What is Emmet and why would we want it?

omar2205:
[Emmet](https://emmet.io/) is a plugin that allows you for example to type `h1.hero-title` and press tab to convert it to `<h1 class="hero-title"></h1>`.

You could also press tab after you write an HTML element name and convert it into a tag. You could generate lorem ipsum text with it, `lorem5` will generate 5 words.

Also, you could generate HTML easily with it, for example: 
`ul>li*5{Item $$$}.item.item--$$$`
will generate:
```html
<ul>
  <li class="item item--001">Item 001</li>
  <li class="item item--002">Item 002</li>
  <li class="item item--003">Item 003</li>
  <li class="item item--004">Item 004</li>
  <li class="item item--005">Item 005</li>
</ul>
```

It makes it easier to build HTML quickly.

benmccann:
thanks, but I don't think the REPL is the right place to include stuff like this. we're just trying to teach users the basics and don't want to make our environment non-standard

omar2205:
I mean, it's being used by most IDEs, but that's okay.

TheBlckbird:
i think this should be reopened as you’re not forced to use it but it’s very useful when you want to create a small project in the repl.

dominikg:
iirc the repl is using codemirror6 now but the emmet plugin is still 35kb gzip https://bundlephobia.com/package/@emmetio/codemirror6-plugin@0.3.0

This would have to be downloaded and executed by every repl user. The repl is not intended to be a full IDE. If you want to have a more IDE like experience, you can use stackblitz via https://vite.new/svelte or https://sveltekit.new.

A nice middleground is https://www.sveltelab.dev/

omar2205:
35kb is acceptable for a nice-to-have feature

PR #263, opened 2022-01-07
update and organize paths and imports

ignatiusmb:
I saw @bluwy's #248 and seems like I missed updating some paths when migrating site-kit to SvelteKit. Hopefully I caught all of them here.


bluwy (approved):
LGTM. Unrelated, but we could probably do a big Prettier format soon to make things consistent too.

benmccann (approved):

PR #270, opened 2022-01-13
Update to latest Kit and handle breaking changes

geoffrich:
This PR updates the repo to the latest version of SvelteKit, and handles the breaking changes around `page`, `query`, and the updated `Request` and `Response`. This is a prerequisite for #245, which requires the change from next-215 to make the skip link work properly.


Conduitry (changes requested):
We should probably keep using the npm tag `next` in all of the `package.json` files for Kit and its adapters.

It also doesn't look like what you're updating to is the latest version anymore.

geoffrich:
Thanks, I made most of these changes yesterday and missed the new releases. Updated to actual latest and put `next` back in the `package.json`

geoffrich:
I'll do some validation on the preview deployments to make sure everything is working properly.

geoffrich:
Looks like this broke anything involving a Vercel serverless function (server rendering and calling endpoints), though everything works fine locally. Going to mark as draft until I can figure out what's going on.

image

geoffrich:
Blocked by https://github.com/sveltejs/kit/issues/3343

sites/svelte.dev/src/routes/auth/callback.js
```diff
@@ -44,7 +44,7 @@ export async function get({ host, query }) {
 					expires: new Date(expires),
 					path: '/',
 					httpOnly: true,
-					secure: !host.startsWith('localhost:')
+					secure: !url.host.startsWith('localhost:')
```
Conduitry, line 47: Would it make sense to use `url.protocol` here (and in the other places we're checking for `localhost`)?

sites/svelte.dev/src/routes/auth/login.js
```diff
@@ -2,15 +2,15 @@ import { stringify } from 'querystring';
...
+	? ({ url }) => {
+			const protocol = url.host.startsWith('localhost:') ? 'http' : 'https';
 
 			const Location =
 				`${oauth}/authorize?` +
 				stringify({
 					scope: 'read:user',
 					client_id,
-					redirect_uri: `${protocol}://${host}/auth/callback`
+					redirect_uri: `${protocol}://${url.host}/auth/callback`
```
Conduitry, line 13: Would it make sense to use `url.origin` here?

sites/svelte.dev/src/routes/auth/callback.js
geoffrich: Looks like `url.protocol` is "http:", but `url.hostname` is "localhost" (for http://localhost:3000). Did you mean the latter?

benmccann: I think he means to change the check to `secure: url.host.startsWith('https:')`, which sounds like a good idea to me

geoffrich: Ah, gotcha. Will update.

geoffrich: See https://github.com/sveltejs/sites/pull/270/commits/9f8a47f6b0201d29fe03d44138660eae507f4524 -- lmk if that's not what you were looking for.

sites/svelte.dev/src/routes/apps/destroy.js
```diff
@@ -1,11 +1,12 @@
 import * as gist from '$lib/db/gist';
 
-export async function post({ locals, body }) {
+export async function post({ locals, request }) {
```
benmccann, line 3: will this need to be `event` now?

sites/svelte.dev/src/routes/repl/create.json.js
```diff
@@ -1,10 +1,11 @@
 import * as gist from '$lib/db/gist';
 
-export async function post({ locals, body }) {
+export async function post({ locals, request }) {
```
benmccann, line 3: will this need to be `event` now?

sites/svelte.dev/src/routes/apps/destroy.js
geoffrich: I think it's correct (though I still need to E2E test). The argument passed post is `event`, but I'm destructuring the request from it. https://kit.svelte.dev/docs#routing-endpoints-body-parsing

sites/svelte.dev/src/routes/repl/create.json.js
geoffrich: [See other comment.](https://github.com/sveltejs/sites/pull/270#discussion_r789086200)

sites/svelte.dev/src/routes/apps/destroy.js
benmccann: ah, right. nevermind

benmccann (approved):

geoffrich:
Waiting to leave draft state until I can E2E test, since this touches all functionality on the site (saving/deleting REPLs, login, etc). @benmccann can you authorize the preview deployment of svelte.dev to Vercel?

benmccann (changes requested):
actually I take back my approval :smile:  since you've updated the `site-kit` code I think you'll also need to upgrade `kit.svelte.dev` to the latest version of SvelteKit as well or else I'm guessing that site will break since it depends on the `workspace` version of `site-kit`:

https://github.com/sveltejs/sites/blob/99870b7457868142c3b5c91d29083a5e3c3f81c4/sites/kit.svelte.dev/package.json#L14

geoffrich:
@benmccann I think I did -- see the [pnpm lockfile](https://github.com/geoffrich/svelte-sites/blob/9f8a47f6b0201d29fe03d44138660eae507f4524/pnpm-lock.yaml#L89). It also seems to work properly when running locally.

Finished testing and everything looks good, so I think this is ready for final approval now. Login stuff didn't work on the preview deployment, but I assume that's because the GitHub OAuth is only set up for svelte.dev. Running locally I was able to login, logout, and save & delete REPLs.

benmccann (approved):
ok, yeah, you're right. I expected we would have at least one endpoint needing to be migrated in either the HN or SvelteKit sites, but I couldn't find any. I guess I forgot they're all in api.svelte.dev, so there's not really anything to do here

PR #272, opened 2022-01-19
add hn.svelte.dev

Rich-Harris:


Conduitry:
Do we want a more personalized readme for this?

Do we want to use the shared `.prettierrc (and maybe other files?) rather than giving this project its own?

benmccann:
Hmm. My initial reaction is that this would cause quite a bit of headache for me. I use the `hn.svelte.dev` example all the time when developing against Kit since it lives in the same repo. Perhaps I could use the `basics` app, but it lives a bit deeply nested within the repository and it doesn't really have an easily navigable homepage showing the various pieces of functionality that exist

Rich-Harris:
good catch, removed the prettierrc and updated the README

PR #273, opened 2022-01-19
update local svelte setup

bluwy:
Update docs and `repl/local` to work with a local copy of Svelte.

### DIFF
--- a/sites/svelte.dev/.env.example
+++ b/sites/svelte.dev/.env.example
@@ -4,7 +4,7 @@ GITHUB_CLIENT_SECRET=
 
 # Path to local copy of Svelte relative from sites/svelte.dev. Used by the REPL.
 # Optional. The default value assumes the svelte repo and sites repo are in the same directory.
-LOCAL_SVELTE_PATH=../../../svelte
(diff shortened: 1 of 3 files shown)


sites/svelte.dev/.env.example
```diff
@@ -2,6 +2,10 @@
 GITHUB_CLIENT_ID=
 GITHUB_CLIENT_SECRET=
 
+# Path to local copy of Svelte relative from sites/svelte.dev. Used by the REPL.
+# Optional. The default value assumes the svelte repo and sites repo are in the same directory.
+LOCAL_SVELTE_PATH=../../../svelte
```
benmccann, line 7: maybe this shouldn't be enabled by default so that you don't have to have Svelte checked out and can just use the published version?
```suggestion
# LOCAL_SVELTE_PATH=../../../svelte
```

bluwy: The env var would only be used if you're explicitly testing with `?version=local`, so it would be fine if it isn't checked out. But I think it makes sense to comment it out too to denote that it's optional.

Conduitry (approved):

PR #277, opened 2022-01-20
Add limits to grid items

mattpilott:
Prevents incorrectly sized items as seen in Safari
Screenshot 2022-01-20 at 8 28 46 pm


bluwy:
Looks like #272 updated some deps in svelte.dev that's causing the build to fail. Presumably should be fix with #270

PR #278, opened 2022-01-22
Fix breaking change not handled for REPL endpoint

Theo-Steiner:
#270 updated to the latest kit version and addressed breaking changes throughout the whole site, but @geoffrich must have overlooked this endpoint in the REPL.
Since saving REPLs is broken by this, I thought it would be fastest if I just straight into a PR.
Following the instructions to set up supabase and github oauth this works like a charm on localhost.
Much love <3 to all of you heroes working on the site


geoffrich (approved):
Thanks! I missed that there was a separate endpoint for editing a REPL 😅

PR #281, opened 2022-01-27
[chore] improve local REPL without local Supabase

Conduitry:
This is a followup to #273 and makes running the svelte.dev site locally a bit easier, for the benefit of folks who want to try out local copies of the compiler on saved REPLs (say, on one given as a repro for an open issue).

In dev mode, you are no longer required to set the `SUPABASE_` env vars, and attempting to load a REPL without them set will instead proxy the request to the production svelte.dev, similar to what happened in the old version of the Svelte site.

This also updates the readme for svelte.dev slightly with appropriate information.

### DIFF
--- a/sites/svelte.dev/README.md
+++ b/sites/svelte.dev/README.md
@@ -1,6 +1,10 @@
 ## Running locally
 
-Setup a database on [Supabase](https://supabase.com) with the instructions [here](../../db) and set the corresponding environment variables.
+A local database is only required in dev mode if you want to test reading and writing saved REPLs on it. Without a local database in dev mode, the REPL will be able to load saved REPLs from the production database, but not save them.
+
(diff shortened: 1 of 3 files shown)


sites/svelte.dev/README.md
```diff
@@ -1,6 +1,10 @@
 ## Running locally
 
-Setup a database on [Supabase](https://supabase.com) with the instructions [here](../../db) and set the corresponding environment variables.
+A local database is only required in dev mode if you want to test reading and writing saved REPLs on it. Without a local database in dev mode, the REPL will be able to load saved REPLs from the production database, but not save them.
+
+Note also that in dev mode, the REPL will currently only work in Chrome, as noted in the Vite documentation [here](https://vitejs.dev/guide/features.html#web-workers), pending support in Firefox for `import` statements in web workers.
```
benmccann, line 5:
```suggestion
Note also that in dev mode, the REPL will currently only work in Chrome, [as noted in the Vite documentation](https://vitejs.dev/guide/features.html#web-workers), pending support in Firefox for `import` statements in web workers.
```

benmccann (approved):

PR #283, opened 2022-01-28
introduce `sites/svelte.dev` to prettier

ignatiusmb:


benmccann:
seems like a good idea to me. happy to merge this if you want to rebase it and actually run the formatting

ignatiusmb:
Will do a follow up for the other packages/sites in a separate PRs and turn on the checks in CI after

bluwy (approved):
LGTM. I wonder if it's worth/possible to run prettier for all workspaces at once, e.g. running formatter in the repo root.

ignatiusmb:
I'm not sure, I might've tried something like that at first and failed with weirdly. Kit runs the formatter for each packages recursively from the root if that's what you mean by running for all at once. With `pnpm -r format`, we just need each packages to have `format` command in the `scripts`

bluwy:
Yeah, I meant having one prettier command in the root that formats the world (not `pnpm -r format`). SvelteKit seems to be splitting it per workspace basis so that it only formats workspace files, and I think that's a fine approach too if we want to follow it.

ignatiusmb:
Sounds like a good idea and probably makes more sense for this repo since every package is a SvelteKit project, I'll retry that approach on the next PR

PR #298, opened 2022-02-09
Add syntax higlighting styles to FAQ

geoffrich:
Followup to https://github.com/sveltejs/sites/pull/297

The code block CSS was not imported on the FAQ page, so the styles weren't applied when loading the page directly.


geoffrich:
@benmccann this should fix the issue you were seeing.

PR #309, opened 2022-03-01
Fix overflow to remove empty scroll-track

MathiasWP:


benmccann:
Can you give more details about the empty scroll such as what pages it appears on with which browsers and possibly a screenshot?

MathiasWP:
Sure! @benmccann 

Screenshot 2022-03-01 at 18 36 31
Screenshot 2022-03-01 at 18 38 18

Screenshot 2022-03-01 at 18 38 34

PR #326, opened 2022-04-12
fix build error with svelte-json-tree

tanhauhau:
i'm not quite sure what happened, but @bluwy suggested to add `"svelte"` entry to `@sveltejs/repl`


bluwy (approved):
Build seems to work now 🎉

PR #327, opened 2022-04-13
Increase the text contrast in the sidebar

TehShrike:
Let us never speak of this again


benmccann:
I feel like this will make it even harder to differentiate the selected item from the others. Maybe we should also bold the selected item?

TehShrike:
Good point.  I pushed a change that bolds the active link (but doesn't bold the hovered/focused links, since those are already getting an extra underline style)

PR #329, opened 2022-04-26
Fix console.assert

nikolai-cc:
Console.assert should only log when its first variable is false. In the current situation it logs to the in-REPL console only in the true case.

I renamed the variable to better reflect what its doing (thinking about it as an assertion rather than a condition hints at its use).

### DIFF
--- a/packages/repl/src/lib/Output/srcdoc/index.html
+++ b/packages/repl/src/lib/Output/srcdoc/index.html
@@ -231,12 +231,12 @@
 			};
 
 			const original_assert = console.assert;
-			console.assert = (condition, ...args) => {
-				if (condition) {
+			console.assert = (assertion, ...args) => {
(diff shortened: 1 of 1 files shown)


MarvinMiles:
Why this is still not merged?

benmccann:
sorry this has gone unreviewed. the site has been rewritten and now lives at https://github.com/sveltejs/svelte.dev. this particular line of code doesn't exist on the new site, so I can't port this PR over and am not sure if the issue still exists. if it does, please send a new PR. the new repo is a bit more active than this one was

nikolai-cc:
Looks like this has been fixed [here](https://github.com/sveltejs/svelte.dev/commit/2977cb7a0eaa3f4cf87dfafd2826a4e5d354eae4)

PR #333, opened 2022-05-04
fix table horizontal scroll

### LINKED ISSUE #332: Documentation is displayed incorrectly in a mobile browser.
Screenshot_2022-05-04-12-42-34-994_com brave browser

bluwy:
Fix #332

Before:
image

After:
image

### DIFF
--- a/packages/site-kit/src/lib/base.css
+++ b/packages/site-kit/src/lib/base.css
@@ -399,7 +399,9 @@ a.no-underline {
 
 /*  tables --------------------------------- */
 table {
+	display: block;
 	width: 100%;
+	overflow-x: auto;
 	font-size: var(--h5);
 }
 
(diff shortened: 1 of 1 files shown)


benmccann:
I'm not sure what I'm supposed to be looking at, but the circled portion below in the after picture looks worse to me

166691114-935b3df2-7b3d-428e-a95d-a4cea662acc3

bluwy:
The screenshots are showing how it looks like on smaller screens (mobile & tablet). If we don't clip the table, it currently causes horizontal scrolling. Maybe we should do it for mobile only? I think horizontal scroll on tablets are still ok.

bluwy:
Aight the screenshots above are in tablet mode, and after another commit it should look like the "Before" image.

But on mobile, the table should clip to prevent horizontal scrolling of the page:

image

(Table is partially scrolled for demonstration)

PR #353, opened 2022-06-24
Fixes REPL targeting bug in Safari

mattpilott:
In Safari, right clicking an element in the REPL preview and selecting "Inspect Element" would yield the wrong iframe as the target.

The correct iframe (.tab-content) has a class of .visible to display it, simple fix is to add `z-index: 1` to the declaration


bluwy:
Tested this on macos Safari, Firefox, Chrome. All seems to work well without side effects. Thanks for the fix!

bluwy (approved):

Issue #288, opened 2022-01-30
[REPL] Splitter does not properly move with pointer on Firefox


andirady:
The pane splitter in the REPL page doesn't follow the cursor on Firefox.
The splitter somehow moved to the left of the screen before following the cursor.

bluwy:
Reproducible in macos m1, firefox 96.0.3

https://user-images.githubusercontent.com/34116392/151751050-24f00c61-64df-47ef-a7e1-fab9ad515e23.mov

brunnerh:
Can also happen in Chrome (v.97.0.4692.99, Win 10, 64bit)

Mlocik97:
can reproduce in firefox, but in chrome it works fine for me.

geoffrich:
I can reproduce on Firefox, but noticed something interesting: once I resize the screen, the splitter moves with the cursor.

andirady:
As @geoffrich mentioned, the splitter behaves correctly after the window was resized.
I ran some debugging and found out that the ``size`` at startup is somehow twice the actual size.
For example, the actual div with is 2560px, but at start up, ``size`` is 5120px as shown in screenshot below:

![image](https://user-images.githubusercontent.com/385785/153720543-14debc3d-537b-4857-af62-899089b93aef.png)

This is something to do with ``bind:clientWidth``. It could be that the bug is at svelte, not on the ``SplitPane.svelte``. 🤔

andirady:
I did some debugging, it seems that at startup, the ``width`` field in ``index.svelte`` is 0. This causes the ``mobile`` field to be set, hence setting the ``$toggleable`` to ``true``. Which enables the ``.toggleable`` class hence the ``size`` in ``SplitPane.svelte`` is computed to be twice the size of the screen.
The ``$toggleable`` immediately turns ``false`` but the ``size`` in ``SplitPane.svelte`` doesn't change, probably since it's already mounted.

bluwy:
Fixed in #359

PR #359, opened 2022-07-19
fix REPL pane resizing

Rich-Harris:
i'm surprised we didn't get yelled at more for this. the REPL is kinda busted at the moment — Firefox doesn't know how big the container should be, and the `<SplitPane>` component keeps losing track of the mouse as soon as it touches the iframe:

https://user-images.githubusercontent.com/1162160/179646274-ba71e685-985e-4e12-81b2-6afdea92eb9d.mov

This fixes both.


bluwy:
We did got yelled at #288 😄

PR #361, opened 2022-08-10
Add Fetch Rewards to the list of Who's Using Svelte

oettingerj:
This adds [Fetch Rewards](https://www.fetchrewards.com) to the Who's Using Svelte section of the svelte.dev site.

I work at Fetch, and we're using Svelte and SvelteKit extensively in both our public-facing website as well as various internal dashboards. We're huge fans of the Svelte ecosystem and would love to show our support!


benmccann:
Thanks, but we're not adding new logos and this really screws up the layout

Screenshot from 2022-08-10 11-05-31

PR #385, opened 2022-10-15
[chore] unpin sveltekit

benmccann:


Conduitry:
I'm seeing errors about `devalue.uneval` not being a function in the Vercel build logs, which means devalue 3 is being called as though it were devalue 4. I see both devalue 3 and 4 in the lockfile - we need to make sure the correct version is being used. I don't think we want 3 around at all.

benmccann:
Thanks for the review. I updated it to use devalue 4 everywhere

Conduitry (changes requested):
Ah, I didn't notice that `devalue` was also a direct dependency of svelte.dev. That might explain why SvelteKit got stuck with the older version.

`sites/svelte.dev/src/routes/auth/callback/+server.js` needs to be updated to use the `uneval` export now, or GitHub auth will not work. (The reason this isn't caught at build time is that there _is_ still a named `devalue` export from the module, but one which throws an error at runtime telling you to use `uneval`.)

Conduitry:
I've just opened https://github.com/Rich-Harris/devalue/pull/41, which would have prevented this build from succeeding.

benmccann:
thanks for catching that. I've made that change

Conduitry (changes requested):
Where `devalue()` is called down lower in the file also needs to be changed.

benmccann:
fixed. I'm too used to relying on the compiler :stuck_out_tongue_closed_eyes: I added `svelte-check` which would have caught this. I'll wait to enforce it on the CI in a new PR because there's too many errors that need to be fixed to make it part of this one

PR #386, opened 2022-10-15
[fix] add missing images

benmccann:


baseballyama:
Do we need to use WebP?

https://developers.google.com/speed/webp#:~:text=WebP%20lossless%20images%20are%2026,of%20just%2022%25%20additional%20bytes.

benmccann:
We should probably make `manifest.json` an endpoint that serves images that have been passed through Vite's image processing pipeline. Then we could serve webp, include hashes in the name and cache them, etc. But for now I just want to fix the 404s by restoring the images that were in https://github.com/sveltejs/hn.svelte.dev

baseballyama (approved):

PR #387, opened 2022-10-17
setup redirects

benmccann:
Serve svelte.dev/roadmap so we can easily host it elsewhere in the future if we wish

The getting started page now lives at https://kit.svelte.dev/docs/introduction#getting-started

### DIFF
--- a/sites/svelte.dev/src/hooks.server.js
+++ b/sites/svelte.dev/src/hooks.server.js
@@ -1,9 +1,0 @@
-export const handle = async ({ event, resolve }) => {
-	if (event.url.pathname === '/blog/the-easiest-way-to-get-started') {
-		return Response.redirect(`${event.url.origin}/docs/introduction`, 308);
-	}
-	if (event.url.pathname === '/roadmap') {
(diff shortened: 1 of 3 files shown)


Conduitry:
Shouldn't we have `src/routes/roadmap/+server.js` etc. and let just SvelteKit do the routing for us?

benmccann:
Would we also have a `blog/the-easiest-way-to-get-started/+page.svelte`?

I don't quite get the benefit of doing it that way. It seemed to me like it'd be nice to have one place for all the redirects to live.

Rich-Harris:
We have precedent in the form of https://github.com/sveltejs/sites/blob/master/sites/svelte.dev/src/routes/chat/%2Bserver.js. Though they should probably be this...

```js
// +page.js
import { redirect } from '@sveltejs/kit';

export function load() {
  throw redirect(307, '...');
}
```

...so that they redirect in the client without a server roundtrip.

You could argue that there's a distinction between 'vanity' routes like `/chat` and `/roadmap` that are actually shortcuts to external sites, and proper redirects like `/blog/easiest...` -> `/docs/introduction` that could be dealt with in `handle`, but I tend to think that individual routes make more sense across the board.

benmccann:
Ok. Updated

Issue #10, opened 2021-03-16
fix <pre> styles


Rich-Harris:
should be horizontally scrollable instead of having overflow:

![image](https://user-images.githubusercontent.com/1162160/111384619-19b28d80-8680-11eb-99dc-ddd0e72c96ca.png)

PR #392, opened 2022-11-01
feat: use optimised S logo

mxdvl:
Minimal visual difference for under half the code.

As proposed [in the branding repo #10](https://github.com/sveltejs/branding/pull/10), this version of the S uses stroked path with covered ends, rather than an outline.

image

The benefit of this approach is that your source assets are [defined as Svelte components](https://svelte.dev/repl/9dff44b01d2642168a56637f03652936?version=3.52.0), which beyond dogfooding also allows freeing these assets from external software or service the source might live. The performance improvements are most likely negligible.


nosovk:
@mxdvl 
I actually don't know how you do it, but looks super
image
It's twice less from original!

mxdvl:
@nosovk good vector mathematics teachers and hours spent in the MDN docs on SVG paths? [The interactive REPL has a debug more highlighting the underlying geometry](https://svelte.dev/repl/9dff44b01d2642168a56637f03652936?version=3.55.1)

benmccann:
I don't know that it's worth optimizing this temporary version of the logo. Assuming the war doesn't last forever, the logo will eventually revert back to the original. Let's consider https://github.com/sveltejs/branding/pull/10 instead and have any discussion about whether to make the change there

PR #395, opened 2022-11-10
[fix] get REPL running on latest Svelte version

dummdidumm:
this is an attempt at a quick fix adding a polyfill of "util" (a node built-in) which is required by Svelte version 3.53 - therefore REPL etc fails in the browser. Afterwards we can think about the proper way to fix this (either inlining the util package in the Svelte compiler or providing that polifyll through the repl package, or keep this here).

Update: might not work since compiler.mjs needs to be made aware of utils somehow, not sure how we can do that.


dummdidumm:
Unfortunately that doesn't work, this all happens inside the worker, which is within the REPL

PR #397, opened 2022-11-11
fix importing from other cdns

sibbng:
Fixes imports from other CDNs like esm.sh and jspm.dev

URL for testing: https://esm.sh/@vueuse/core@9.5.0/


benmccann:
thank you! sorry this PR has gone unreviewed. could you send it to https://github.com/sveltejs/svelte.dev instead where the site and repl currently live and which is a bit more active?

PR #398, opened 2022-11-18
Fix REPL 404 handling

benmccann:
Ref https://github.com/sveltejs/sites/issues/399

### DIFF
--- a/sites/svelte.dev/src/lib/db/gist.js
+++ b/sites/svelte.dev/src/lib/db/gist.js
@@ -29,7 +29,7 @@ export async function list(user, { offset, search }) {
 /**
  * @param {User} user
  * @param {Pick<Gist, 'name'|'files'>} gist
- * @returns {Gist}
+ * @returns {Promise<Gist>}
  */
 export async function create(user, gist) {
 	const { data, error } = await client.rpc('gist_create', {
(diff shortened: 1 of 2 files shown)


sites/svelte.dev/src/lib/db/gist.js
```diff
@@ -83,7 +83,7 @@ export async function update(user, gistid, gist) {
 
 /**
  * @param {number} userid
- * @param {string[]} ids
+ * @param {Promise<string[]>} ids
```
Rich-Harris, line 86: this seems wrong

sites/svelte.dev/src/routes/(authed)/repl/[id].json/+server.js
```diff
@@ -43,7 +43,10 @@ export async function GET({ params }) {
 		const res = await fetch(`${API_BASE}/docs/svelte/examples/${params.id}`);
 
 		if (!res.ok) {
-			return new Response(await res.json(), { status: res.status });
+			return new Response(await res.json(), {
+				status: res.status,
+				headers: { 'Content-Type': 'application/json' }
+			});
```
Comment (Rich-Harris) [inline sites/svelte.dev/src/routes/(authed)/repl/[id].json/+server.js:49]: not sure i understand why this isn't `throw error(res.status, await res.json())` or even just `throw error(res.status)`?

Comment (benmccann) [reply sites/svelte.dev/src/routes/(authed)/repl/[id].json/+server.js:49]: the second param to `error` must be an `App.Error`. I don't know exactly what `res.json()` would contain and if it would adhere to that type

PR #403, opened 2022-11-21
this was erroneously converted to a promise

Rich-Harris:


sites/svelte.dev/src/lib/db/gist.js
```diff
@@ -83,7 +83,7 @@ export async function update(user, gistid, gist) {
 
 /**
  * @param {number} userid
- * @param {Promise<string[]>} ids
+ * @param {string[]} ids
```
benmccann, line 86: Doesn't the return type have to be a Promise if the function is async? VS Code was complaining on the other functions that we were doing an unnecessary `await` because of how the return type was documented. I didn't check this one specifically but assume it would be the same

Conduitry: This is the `ids` param, not the return value.

benmccann: oh, oops. I just looked that it was last

PR #405, opened 2022-11-21
upgrade supabase client

benmccann:
migration guide here: https://supabase.com/blog/supabase-js-v2

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -38,7 +38,7 @@ importers:
       yootils: 0.3.1
     devDependencies:
       '@sveltejs/adapter-auto': 1.0.0-next.89
-      '@sveltejs/kit': 1.0.0-next.554_svelte@3.53.1+vite@3.2.4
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
```diff
@@ -38,7 +38,7 @@ importers:
       yootils: 0.3.1
     devDependencies:
       '@sveltejs/adapter-auto': 1.0.0-next.89
-      '@sveltejs/kit': 1.0.0-next.554_svelte@3.53.1+vite@3.2.4
+      '@sveltejs/kit': 1.0.0-next.555_svelte@3.53.1+vite@3.2.4
```
dominikg, line 41: was this on purpose or accidental because kit version is `next` in package.json?

benmccann: I guess pnpm just did that automatically. There shouldn't be any harm in it though

Rich-Harris (approved):

PR #410, opened 2022-12-07
recommend sveltekit on homepage

benmccann:


Conduitry:
Do we need to do something here to tell people that they'll have to interact with prompts now? I'm not sure what that would look like, but the experience _has_ changed.

Do we want part of the command to be a link to some repo, like we did before with the Vite template?

benmccann:
Yeah, it's a good question. For now I was just going to punt on that question because I don't really know how to integrate anything else into the design. Open to ideas though...

PR #413, opened 2022-12-14
prerender `/blog` from markdown in this repo

Rich-Harris:
Currently, content on https://svelte.dev is served via an API that sits in front of a Cloudflare KV store that is populated by a GitHub action that runs when you merge stuff to `master` on `sveltejs/svelte`.

This is absolute madness. Among other things, it makes previewing stuff locally impossible (to a first approximation). We're also leaving performance on the table by not prerendering this stuff.

I'd like to move all the content into this repo, but for now I just did blog posts as that's what I'm currently concerned with. The markdown processing logic should eventually be moved to `site-kit`.


dummdidumm (approved):
I didn't look closely into the markdown transformation code, so I'll ask instead: Where is it from?
Looked at the deployed preview, looks good.

Rich-Harris:
It's a heavily simplified version of the code that powers the kit.svelte.dev docs (simplified because we don't need shiki-twoslash and all that jazz)

PR #417, opened 2022-12-15
chore(blog): change final pre-release to 589

theetrain:
Not sure if intentional to have it as 587 when the final pre-release is 589.


Conduitry:
It should maybe actually be 588. It should not be 589, because that removes all of the deprecation warnings and errors in preparation for 1.0.

theetrain:
Fair point, I see v588 mentioned in the revised migration guide: https://github.com/sveltejs/kit/discussions/5774

It's now referencing 588. 👍

PR #438, opened 2023-02-10
pin site versions

dummdidumm:
- pin repl version so we can freely adjust it in the meantime (it has incompatible style changes)
- make hn/svelte.dev packages private so we can't accidentally publish them

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -93,7 +93,7 @@ importers:
       '@supabase/supabase-js': ^2.1.0
       '@sveltejs/adapter-auto': ^2.0.0
       '@sveltejs/kit': ^1.5.0
-      '@sveltejs/repl': 0.0.3
+      '@sveltejs/repl': 0.0.1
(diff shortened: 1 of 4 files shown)


dummdidumm:
I don't understand the mysterious pnpm lock failure

dummdidumm:
This is getting weirder by the minute. I removed all node_modules and the lockfile locally to start a fresh `pnpm i` and the error persists

dummdidumm:
From tests in #439 it seems that adding back `publishDirectory: package` to the lockfile fixes the issue. No idea why

dummdidumm:
ah yeah that's a bummer. We can't update those packages right now because they work with the new style guide now - we plan to release an overhaul of the Svelte site soon, it should be in there then.

Tal500:
> ah yeah that's a bummer. We can't update those packages right now because they work with the new style guide now - we plan to release an overhaul of the Svelte site soon, it should be in there then.

Any update?

benmccann:
We'll release the new site in the next week or two

PR #445, opened 2023-02-25
feat: add support for `exports` field to REPL

### LINKED ISSUE #134: REPL module resolution doesn't respect "exports" entry in package.json.
The REPL's node module resolution doesn't respect the "exports" entry in packages' package.json files. This could mean that the REPL is running on an old version of node that didn't support this, but I don't know how to tell.

You can see this behavior here:
https://svelte.dev/repl/f9c535598baa4694b57f3d47dab4e58e?version=3.44.1

The error reads:
`Cannot find "/index.js" in @smui/textfield@6.0.0-beta.0`

The package.json for @smui/textfield@6.0.0-beta.0 includes this entry:

```json
  "exports": {
    "./package.json": "./package.json",
    ".": "./dist/index.js",
    "./character-counter": "./dist/character-counter/index.js",
    "./helper-text": "./dist/helper-text/index.js",
    "./icon": "./dist/icon/index.js"
  },
```

Tal500:
Finally fixes #134.
Based on https://github.com/sveltejs/sites/issues/134#issuecomment-995449952 and https://github.com/sveltejs/sites/issues/134#issuecomment-1444393511.

~Still on draft since~ I don't know how to report the errors in importing to the user, see the two "TODO"s.

Related context: https://github.com/benmccann/esm-env/pull/2#issuecomment-1444437790

### DIFF
--- a/packages/repl/src/lib/workers/bundler/index.js
+++ b/packages/repl/src/lib/workers/bundler/index.js
@@ -110,6 +110,7 @@ function has_loopGuardTimeout_feature() {
 async function get_bundle(uid, mode, cache, lookup) {
 	let bundle;
 
+	/** A set of package names (without subpaths) to include in pkg.devDependencies when downloading an app */
 	const imports = new Set();
 	const warnings = [];
 	const all_warnings = [];
--- a/packages/repl/src/lib/workers/bundler/index.js
+++ b/packages/repl/src/lib/workers/bundler/index.js
@@ -179,7 +179,7 @@ async function get_bundle(uid, mode, cache, lookup) {
 							exprots_resolver(pkg, importee, { browser: true, conditions: ['production'] }) ??
 							legacy_resolver(pkg, {
 								browser: importee,
-								fields: [['svelte', 'browser', 'module', 'main']]
+								fields: ['svelte', 'browser', 'module', 'main']
 							});
 
 						if (resolved_id === false) {
--- a/packages/repl/src/lib/workers/bundler/index.js
+++ b/packages/repl/src/lib/workers/bundler/index.js
@@ -175,7 +175,7 @@ async function get_bundle(uid, mode, cache, lookup) {
 						const pkg = JSON.parse(pkg_json);
 
 						/** @type {string | false | undefined} */
-						const resolvedId =
+						const resolved_id =
 							exprotsResolver(pkg, importee, { browser: true, conditions: ['production'] }) ??
 							legacyResolver(pkg, {
 								browser: importee,
--- a/packages/repl/src/lib/workers/bundler/index.js
+++ b/packages/repl/src/lib/workers/bundler/index.js
@@ -160,7 +160,7 @@ async function get_bundle(uid, mode, cache, lookup) {
 
 				if (importee_package_name_match) {
 					const importee_package_name = importee_package_name_match[0];
-					
+
 					if (importer in lookup) {
 						imports.add(importee_package_name);
 					}
@@ -173,9 +173,14 @@ async function get_bundle(uid, mode, cache, lookup) {
 						const pkg_json = (await fetch_if_uncached(pkg_url, uid)).body;
 						const pkg = JSON.parse(pkg_json);
 
+						const pkg_url_base = pkg_url.replace(/\/package\.json$/, '');
+
 						/** @type {string | false | undefined} */
 						const resolved_id =
-							exports_resolver(pkg, importee, { browser: true, conditions: ['svelte', 'production'] }) ??
+							exports_resolver(pkg, importee, {
+								browser: true,
+								conditions: ['svelte', 'production']
+							}) ??
 							legacy_resolver(pkg, {
 								browser: importee,
 								fields: ['svelte', 'browser', 'module', 'main']
@@ -183,9 +188,24 @@ async function get_bundle(uid, mode, cache, lookup) {
 
 						if (resolved_id === false) {
 							// TODO: Output an error to the user that the package author doesn't want the user to import this path
-						} else if (resolved_id != null) {
-							const url = pkg_url.replace(/\/package\.json$/, '');
-							return new URL(resolved_id, `${url}/`).href;
+						} else if (resolved_id == null) {
+							// If not resolved, there is still hope by the NPM standards to resolve it
+							//  to the file `./index.{mjs|js}`, if exists.
+							if (importee_package_name === importee) {
+								// Return the first one that doesn't throw, if any
+								for (const ext of ['.mjs', '.js']) {
+									try {
+										const indexUrl = new URL(`index.${ext}`, `${pkg_url_base}/`).href;
+										return (await fetch_if_uncached(indexUrl, uid)).url;
+									} catch {
+										// ignore, try the next option
+									}
+								}
+
+								// TODO: Throw an error that there no main entry field were found nor './index.mjs' or './index.js' file exist.
+							}
+						} else {
+							return new URL(resolved_id, `${pkg_url_base}/`).href;
 						}
 					} catch (err) {
 						// ignore
--- a/packages/repl/src/lib/workers/bundler/index.js
+++ b/packages/repl/src/lib/workers/bundler/index.js
@@ -165,51 +165,88 @@ async function get_bundle(uid, mode, cache, lookup) {
 						imports.add(importee_package_name);
 					}
 
+					const fetch_package_info = async () => {
+						try {
+							const pkg_url = await follow_redirects(
+								`${packagesUrl}/${importee_package_name}/package.json`,
+								uid
+							);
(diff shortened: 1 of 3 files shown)


packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -1,4 +1,5 @@
 import { rollup } from '@rollup/browser';
+import { exports as exprotsResolver, legacy as legacyResolver } from 'resolve.exports';
```
benmccann, line 2: typo in "exports" and should probably use `_` rather than camelCase as it's an internal variable
```suggestion
import { exports as exports_resolver, legacy as legacy_resolver } from 'resolve.exports';
```

packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -155,22 +156,42 @@ async function get_bundle(uid, mode, cache, lookup) {
...
+					try {
+						const pkg_url = await follow_redirects(
+							`${packagesUrl}/${importee_package_name}/package.json`,
+							uid
+						);
+						const pkg_json = (await fetch_if_uncached(pkg_url, uid)).body;
+						const pkg = JSON.parse(pkg_json);
+
+						/** @type {string | false | undefined} */
+						const resolvedId =
```
benmccann, line 178:
```suggestion
						const resolved_id =
```

packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -155,22 +156,42 @@ async function get_bundle(uid, mode, cache, lookup) {
...
+							exprotsResolver(pkg, importee, { browser: true, conditions: ['production'] }) ??
```
benmccann, line 179: we should use the `svelte` condition as well
```suggestion
							exports_resolver(pkg, importee, { browser: true, conditions: ['svelte', 'production'] }) ??
```

Tal500: Correct me if I'm wrong, but the `svelte` condition is obsolete and isn't recommended anymore, nor extended to the `exports` field

packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -155,22 +156,42 @@ async function get_bundle(uid, mode, cache, lookup) {
...
+						);
+						const pkg_json = (await fetch_if_uncached(pkg_url, uid)).body;
+						const pkg = JSON.parse(pkg_json);
+
+						/** @type {string | false | undefined} */
+						const resolved_id =
+							exprots_resolver(pkg, importee, { browser: true, conditions: ['production'] }) ??
+							legacy_resolver(pkg, {
+								browser: importee,
+								fields: [['svelte', 'browser', 'module', 'main']]
```
benmccann, line 182: I don't think this should be an array of arrays looking at the examples on https://github.com/lukeed/resolve.exports
```suggestion
								fields: ['svelte', 'browser', 'module', 'main']
```

Tal500: Woops! Totally right

packages/repl/src/lib/workers/bundler/index.js
benmccann: Nope, the reverse. It's something we've just recently started promoting. The `svelte` field in the legacy resolver is less recommended now as we're recommending people to use the `svelte` condition instead, but I'd still continue to support it here

packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -1,4 +1,5 @@
 import { rollup } from '@rollup/browser';
+import { exports as exprots_resolver, legacy as legacy_resolver } from 'resolve.exports';
```
benmccann, line 2: the typo is still there :laughing: 
```suggestion
import { exports as exports_resolver, legacy as legacy_resolver } from 'resolve.exports';
```

Tal500:
I have a question of the Svelte plugin behavior - Should there will be separation of import behavior between "*.svelte" files and "*.js" files? Specifically:
1. Should the "svelte" field directly in "package.json" or nested inside "externals" be respected only when the importer is a "*.svelte" file? (and not for a "*.js" file)
2. I understand that Svelte forbid the user to use the "require" keyword, but rather forces him to use the "import" keyword. Should we allow the "require" keyword for "*.js" files(or is it currently being supported)? If so, then we need to check if Rollup passed the import with the query "?require=true", and pass the "import" flag to the `externals_resolver` function. (Should we in general check other Rollup import query flags in `resolveId()`? Are there other flags even at all?)

Tal500:
@benmccann How should errors from imports be reported to the REPL user?

benmccann (approved):
> @benmccann How should errors from imports be reported to the REPL user?

I don't know, but that was an existing issue, so I wouldn't let it block this

Tal500:
> I have a question about the Svelte plugin behavior - Should there will be any separation of import behavior between `*.svelte` files and `*.js` files? Specifically:
> 
>     1. Should the "svelte" field directly in "package.json" or nested inside "externals" be respected only when the importer is a `*.svelte` file? (and not for a `*.js` file)
> 
>     2. I understand that Svelte forbid the user to use the "require" keyword, but rather forces him to use the "import" keyword. Should we allow the "require" keyword for `*.js` files(or is it currently being supported)? If so, then we need to check if the user used "require", and pass the "require" flag to the `externals_resolver` function.

Last thing, @benmccann you might missed these question, especially the second one that is more important, to handle the "svelte" record correctly.

benmccann:
> Should there will be any separation of import behavior between *.svelte files and *.js files?

No, I don't think so

> Should we allow the "require" keyword for *.js files(or is it currently being supported)? 

I think we probably don't need to support that in the REPL

Tal500:
> > Should there will be any separation of import behavior between *.svelte files and *.js files?
> 
> No, I don't think so

Thanks! So it makes me wonder - what is the purpose of the "svelte" record in "external"/"package.json"? I thought originally that it was to have a different behaviour for imports inside a svelte file, but now I see it's in general when the build involves Svelte somehow.
So what is the purpose and what does it do? Do you have any "modern" example?

benmccann:
You can use the `svelte` field or export condition to point to Svelte source code and `main` to point to a compiled version if you want to distribute a compiled version for use in non-Svelte projects

Rich-Harris:
Thank you! This is great. I know that only one of the TODOs is new, but I think it would be a good idea to deal with them now otherwise they'll probably still be here in a year's time — I think we just need to throw an appropriately worded error?

Tal500:
> Thank you! This is great. I know that only one of the TODOs is new, but I think it would be a good idea to deal with them now otherwise they'll probably still be here in a year's time — I think we just need to throw an appropriately worded error?

You mean to throw a simple primitive `string`? Will the worker communication magic will display the error string straight to the user?

Tal500:
> Thank you! This is great. I know that only one of the TODOs is new, but I think it would be a good idea to deal with them now otherwise they'll probably still be here in a year's time — I think we just need to throw an appropriately worded error?

Did it right now, possibly by bad English.

Additionally, now in the case where the main entry is requested but doesn't appear, the path that will be taken will be "./index.mjs" or "./index.js" (and error if nothing exists).

packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -107,9 +107,63 @@ function has_loopGuardTimeout_feature() {
...
+			}
+
+			throw `could not find entry point in "${pkg_name}/package.json"`;
+		}
+
+		return resolved_id;
+	}
+
+	if (typeof pkg.browser === 'object') {
+		// this will either return `pkg.browser[subpath]` or `subpath`
```
Tal500, line 154: @Rich-Harris How did you understand by the [browser option docs](https://github.com/lukeed/resolve.exports#optionsbrowser-1) that if no match on "browser", then `resolve.legacy(pkg, { browser: subpath });` is a relative string path?
I think that in this case it will be `undefined`.

Rich-Harris: You can try it in the Node REPL:

```js
resolve = require('resolve.exports');
resolve.legacy(
  {},
  { browser: './foo.js' }
); // undefined

resolve.legacy(
  { browser: {} },
  { browser: './foo.js' }
); // './foo.js'

resolve.legacy(
  { browser: { './foo.js': './bar.js' } },
  { browser: './foo.js' }
); // './bar.js'
```

Tal500: Make sense, knowing this behavior the docs are now less ambiguous on a second glance.

Rich-Harris (approved):

Rich-Harris:
thanks! released 0.2.0

Tal500:
> thanks! released 0.2.0

And I thought that the SvelteKit official releasing process is slow😅

PR #461, opened 2023-04-14
feat(repl): Make SvelteKit agnostic

PuruVJ:
Should not depend on SvelteKit or Vite anymore

svelte-package is updated

No breaking changes

### DIFF
--- a/packages/repl/package.json
+++ b/packages/repl/package.json
@@ -51,6 +51,13 @@
 			"default": "./dist/index.js"
 		}
 	},
+	"typesVersions": {
+		"*": {
+			"*": [
+				"./dist/index.d.ts"
+			]
+		}
+	},
 	"files": [
 		"dist"
 	]
--- a/packages/repl/package.json
+++ b/packages/repl/package.json
@@ -52,7 +52,7 @@
 		}
 	},
 	"typesVersions": {
-		"*": {
+		">4.0": {
 			"*": [
 				"./dist/index.d.ts"
 			]
--- a/packages/repl/package.json
+++ b/packages/repl/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@sveltejs/repl",
-	"version": "0.3.0",
+	"version": "0.2.0",
 	"scripts": {
 		"dev": "vite dev",
 		"build": "vite build",
--- a/packages/repl/package.json
+++ b/packages/repl/package.json
@@ -1,42 +1,57 @@
 {
 	"name": "@sveltejs/repl",
-	"version": "0.2.0",
+	"version": "0.3.0",
 	"scripts": {
 		"dev": "vite dev",
 		"build": "vite build",
 		"package": "svelte-package",
+		"watch": "svelte-package --watch",
 		"preview": "vite preview",
 		"lint": "prettier --ignore-path .gitignore --check --plugin-search-dir=. . && eslint --ignore-path .gitignore .",
 		"format": "prettier --ignore-path .gitignore --write --plugin-search-dir=. ."
 	},
 	"devDependencies": {
 		"@sveltejs/adapter-auto": "^2.0.0",
-		"@sveltejs/kit": "^1.5.2",
-		"eslint": "^8.33.0",
-		"eslint-config-prettier": "^8.6.0",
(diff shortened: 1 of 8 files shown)


packages/repl/package.json
```diff
@@ -5,38 +5,60 @@
...
+	},
+	"exports": {
+		"./package.json": "./package.json",
+		".": {
+			"types": "./dist/index.d.ts",
+			"svelte": "./dist/index.js",
+			"default": "./dist/index.js"
+		}
+	},
+	"typesVersions": {
```
benmccann, line 54: might not need this?

packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -339,8 +340,7 @@ async function get_bundle(uid, mode, cache, lookup) {
 }
 
 async function bundle({ uid, components }) {
-	if (import.meta.env.PROD) {
-		console.clear();
+	if (BROWSER) {
```
benmccann, line 343: should this be `!DEV`? I haven't looked at what it's actually doing, but that would match what it was before

packages/repl/svelte.config.js
```diff
@@ -5,7 +5,4 @@ export default {
 	kit: {
 		adapter: adapter()
 	},
-	package: {
-		exports: (file) => file === 'index.js'
-	}
-};
+	};
```
benmccann, line 8: not sure if the semicolon is required? can definitely remove a couple spaces though
```suggestion
};
```

packages/repl/package.json
PuruVJ: For people using with TS 4, still might need it

packages/repl/src/lib/workers/bundler/index.js
PuruVJ: Oops. Mistake 😅

pnpm-lock.yaml
```diff
@@ -4273,7 +4692,7 @@ packages:
       - rollup
     dev: true
 
-  /vite/4.1.1:
+  /vite@4.1.1:
```
benmccann, line 4695: would be nice to use 4.2.1 everywhere so that we don't have multiple versions

pnpm-lock.yaml
```diff
@@ -4146,72 +4553,78 @@ packages:
...
 
-  /typedarray-to-buffer/3.1.5:
+  /typedarray-to-buffer@3.1.5:
     resolution: {integrity: sha512-zdu8XMNEDepKKR+XYOXAVPtWui0ly0NtohUscw+UmaHiAWT8hrV1rr//H6V+0DvJ3OQ19S979M0laLfX8rm82Q==}
     dependencies:
       is-typedarray: 1.0.0
     dev: false
 
-  /typescript/4.9.5:
+  /typescript@4.9.5:
```
benmccann, line 4611: would be nice to use 5.0.4 everywhere so that we don't have multiple versions

pnpm-lock.yaml
```diff
@@ -4045,19 +4433,40 @@ packages:
...
+      svelte: ^3.55
+      typescript: ^4.9.4 || ^5.0.0
+    dependencies:
+      dedent-js: 1.0.1
+      pascal-case: 3.1.2
+      svelte: 3.58.0
+      typescript: 5.0.4
+    dev: true
+
+  /svelte@3.55.1:
```
benmccann, line 4460: looks like we've got multipe versions of svelte as well

pnpm-lock.yaml
PuruVJ: Other one is site-kit, which is better to not touch IMO. Let's leave it like this for a while IG? Then if we move site-kit back here, it should not be an issue

pnpm-lock.yaml
PuruVJ: Hmm will try updating some deps

benmccann (approved):

PuruVJ:
Ok where is Changeset bot 🤔

PuruVJ:
Merging, as it seems Changeset bot doesn't like commenting in this repo

PR #464, opened 2023-04-15
fix(repl): Compat with latest site-kit

PuruVJ:
Until now some REPL styles were broken. This one changes all CSS variables in REPL to use latest site-kit tokens.

This requires site-kit v5 https://github.com/sveltejs/site-kit/pull/100


packages/repl/src/lib/Output/srcdoc/index.html
```diff
@@ -151,7 +151,7 @@
 				window.addEventListener('unhandledrejection', (event) => {
 					parent.postMessage({ action: 'unhandledrejection', value: event.reason }, '*');
 				});
-			}.call(this));
+			}).call(this);
```
PuruVJ, line 154: I don't understand how the previous syntax worked, and am scared this will break.

PuruVJ: I'm just gonna trust prettier, it never fails me.

PR #483, opened 2023-05-23
revamp(repl): Isolate each REPL instance

PuruVJ:


packages/repl/src/lib/Output/Output.svelte
```diff
@@ -84,6 +84,8 @@
 		ast = compiled.ast;
 	}
 
+	const { module_editor } = /** @type {import('../types').ReplContext}*/ (getContext('REPL'));
```
dummdidumm, line 87: To avoid type casts, repetition of a string etc I suggest to move the set/getContext into a separate file like
```js
// repl-context.js
let key = {}
/** @returns {..} */
export function getREPLContext() {
  return getContext(key);
}
/** @param {..} */
export function setREPLContext(value) {
  setContext(key, value);
}
```

PuruVJ: If it's in its own file, I'm scared if it will undo the isolation. As in, if I have multiple <Repl>, won't they all have access to this same getContext in this case?

PuruVJ: Ok this works. I still don't understand how, but it works 😅

PR #488, opened 2023-06-05
Display full post contents in RSS

sebastiandedeyne:
Instead of only rendering the description, render full post contents in RSS so announcements can be read in RSS readers.

I had a hard time setting up the monorepo locally, I was going to verify the changes on the Vercel preview but need to be authorized first.


benmccann:
The site has been moved to a new location: https://github.com/sveltejs/svelte/tree/master/sites/svelte.dev

PR #489, opened 2023-06-07
feat(repl): v4 support

gtm-nayan:


packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -434,7 +447,7 @@ async function bundle({ uid, files }) {
 				format: 'iife',
 				name: 'SvelteComponent',
 				exports: 'named',
-				sourcemap: true
+				sourcemap: 'inline'
```
benmccann, line 450: will this show up in the JS Output tab? I don't know that I'd want it to. Can we leave it as `true` or should we set it to `false`?

packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -32,20 +32,29 @@ self.addEventListener(
...
 
-				try {
-					importScripts(`${svelte_url}/compiler.js`);
-				} catch {
-					self.svelte = await import(/* @vite-ignore */ `${svelte_url}/compiler.mjs`);
+				const { version } = await fetch(`${svelte_url}/package.json`).then((r) => r.json());
+				console.log(`Using Svelte compiler version ${version}`);
+
+				if (version.startsWith('4')) {
+					// unpkg doesn't set the correct MIME type for .cjs files
```
benmccann, line 42:
```suggestion
					// unpkg doesn't set the correct MIME type for .cjs files
					// https://github.com/mjackson/unpkg/issues/355
```

packages/repl/src/lib/workers/compiler/index.js
```diff
@@ -17,10 +17,19 @@ self.addEventListener(
...
 			case 'init':
-				try {
-					importScripts(`${event.data.svelte_url}/compiler.js`);
-				} catch (e) {
-					self.svelte = await import(/* @vite-ignore */ `${event.data.svelte_url}/compiler.mjs`);
+				const { svelte_url } = event.data;
+				const { version } = await fetch(`${svelte_url}/package.json`).then((r) => r.json());
+
+				if (version.startsWith('4')) {
+					// unpkg doesn't set the correct MIME type for .cjs files
```
benmccann, line 24:
```suggestion
					// unpkg doesn't set the correct MIME type for .cjs files
					// https://github.com/mjackson/unpkg/issues/355
```

packages/repl/src/lib/workers/patch_window.js
```diff
@@ -0,0 +1 @@
+self.window = self; // hack for magic-sring and rollup inline sourcemaps
```
benmccann, line 1:
```suggestion
self.window = self; // hack for magic-sring and rollup inline sourcemaps

```

packages/repl/src/lib/workers/bundler/index.js
gtm-nayan: No. What's shown in the output tab is generated by the other worker, `compiler/index.js`.

PuruVJ (approved):

PR #501, opened 2023-06-27
(feat) add vim keybindings to REPL component

paoloricciuti:
This add support for vim keybings to the REPL component.

A couple of notes:

- Given this ( https://github.com/replit/codemirror-vim/issues/130 ) the vim extension get's resetted upon swapping the state. This means that changing the tab, clicking the errors/warnings or renaming a file will reset the vim extension to the original state (this means that if you are in insert mode it will revert back to normal mode). This is a minor inconvenience that we can't really control since it's up to the mantainers of codemirror-vim to update their state to use facets so that we can serialize the state and store it. If you read the issue the mantainer said that he's working on this use case but it could take a while.
- This only adds the external parameter vim to the repl component, after this pr get's merged i will make another PR to the svelte repo to actually make vim mode work based on a query param (we talked about this with @PuruVJ and given that the repl could see a redesign soon we agreed that for the moment a query param should suffice).


benmccann:
I don't know what the other maintainers think, but it seems odd to me to build support for a particular editor's key bindings into the REPL. Will we also have to add support for emacs, nano, VS Code, etc., etc.?

paoloricciuti:
> I don't know what the other maintainers think, but it seems odd to me to build support for a particular editor's key bindings into the REPL. Will we also have to add support for emacs, nano, VS Code, etc., etc.?

That's fair...i think the reason behing this was because vim users are avid keybings users and a lot of people requested the ability to use vim keybindings in the repl (just like it is on learn.svelte.dev)

PuruVJ:
> I don't know what the other maintainers think, but it seems odd to me to build support for a particular editor's key bindings into the REPL. Will we also have to add support for emacs, nano, VS Code, etc., etc.?

Id say that argument is down the drain since rich added vim mode to learn.svelte.dev 😅

dominikg:
how much weight is this going to add and is there a way to have this as opt-in/lazy loaded feature?

according to bundle-phobia, the codemirror bindings package is 35kb gzip https://bundlephobia.com/package/@replit/codemirror-vim@6.0.14

paoloricciuti:
> how much weight is this going to add and is there a way to have this as opt-in/lazy loaded feature?
> 
> according to bundle-phobia, the codemirror bindings package is 35kb gzip https://bundlephobia.com/package/@replit/codemirror-vim@6.0.14

It is lazy loaded

PR #506, opened 2023-06-28
(fix) loading files from repl json broke the REPL upon renaming

paoloricciuti:
When loading the files from a saved REPL renaming a file broke the REPL because selecting a new tab was storing the state from App.svelte for the newly selected file in the EDITOR_STATE_MAP. Given that the state includes the document itself after renaming the file the editor restored the state with the wrong doc.

See this REPL as an example, trying to rename ItemTypeOne substitute the content of ItemTypeOne.svelte with the one from App.svelte
https://svelte.dev/repl/d9d167fa8ded478ba4dea62257d75a10?version=3.8.0

### DIFF
--- a/packages/repl/src/lib/Repl.svelte
+++ b/packages/repl/src/lib/Repl.svelte
@@ -1,7 +1,8 @@
 <script>
+	import { EditorState } from '@codemirror/state';
 	import { SplitPane } from '@rich_harris/svelte-split-pane';
 	import { BROWSER } from 'esm-env';
-	import { createEventDispatcher, tick } from 'svelte';
+	import { createEventDispatcher } from 'svelte';
 	import { derived, writable } from 'svelte/store';
 	import Bundler from './Bundler.js';
 	import ComponentSelector from './Input/ComponentSelector.svelte';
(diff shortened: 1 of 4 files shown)


packages/repl/src/lib/Repl.svelte
```diff
@@ -50,9 +51,18 @@
 
 		injectedCSS = data.css || '';
 
-		await sleep(50);
```
paoloricciuti, line 53: Here we don't need to sleep anymore...this was to allow the editor to get it's state before setting the editor state map but since we are loading every file with the new state it's not needed anymore

packages/repl/src/lib/Repl.svelte
```diff
@@ -196,8 +206,6 @@
 			$module_editor?.clearEditorState();
 		}
 
-		EDITOR_STATE_MAP.set(filename, $module_editor?.getEditorState());
```
paoloricciuti, line 199: Also here we don't need this anymore because every file get's his entry in the EDITOR_STATE_MAP either when is set at the beginning or upon creation/modification

PuruVJ: Amazing!!

PR #510, opened 2023-06-29
chore: fix worker minification

gtm-nayan:


packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -42,7 +42,7 @@ self.addEventListener(
 					// unpkg doesn't set the correct MIME type for .cjs files
 					// https://github.com/mjackson/unpkg/issues/355
 					const compiler = await fetch(`${svelte_url}/compiler.cjs`).then((r) => r.text());
-					eval(compiler);
+					(0, eval)(compiler);
```
PuruVJ, line 45: Why does this witchcraft even work? 😅

gtm-nayan: https://esbuild.github.io/content-types/#direct-eval

PuruVJ: Naice!

PR #513, opened 2023-06-30
Fix attribute deletion

### LINKED ISSUE #496: REPL breaks when deleting attributes
### Describe the bug

The REPL breaks while deleting whole words (<kbd>CTRL</kbd> + <kbd>Backspace</kbd>) on an element's attributes. 

I've been able to consistently get it to error when you start deleting from an attribute that's assigned to a string literal (.e.g., `value="press me"` in this case).

gif example of the issue

### Reproduction

https://svelte.dev/repl/8abeb99dab034b93a4522a0ba80b1a19?version=4.0.0

### Logs

_No response_

### System Info

```shell
Issue occurs on Chrome
```

### Severity

annoyance

paoloricciuti:
closes #496

This issue is due to the fact that the array of the diagnostics get's out of sync with the actual code for a brief moment. This usually it's not a huge problem because it get's recomputed moments later but it can be a problem when the error falls outside of the max length of source code.

To avoid this issue i've added a bundling promise inside the context that get's reassigned whenever a bundling process starts and resolved immediately after the new bundle has been assigned to the store. The diagnostic has also been transformed from an array that get's recalculated dynamically to an async function that awaits the bundling process. Finally instead of simply passing the diagnostics to neocodemirror we create a `linter ` extension passing the aforementioned async function. Codemirror will call and await this function at each modification.

A couple of points:

1. This means the diagnostics feel a bit more slow because it has to wait for the bundling to finish, however this is more correct since the diagnostics immediately reflect the actual source code.
2. This could also be "fixed" upstream in @PuruVJ @neocodemirror library, right now does work as intended but it uses setDiagnostics and it might be beneficial to use the linter extension instead to allow for async diagnostic calculation. So if @PuruVJ prefer to first upgrade @neocodemirror i can update this pr whenever he push a new release there.

### DIFF
--- a/.changeset/sweet-pianos-sniff.md
+++ b/.changeset/sweet-pianos-sniff.md
@@ -2,4 +2,4 @@
 '@sveltejs/repl': patch
 ---
 
-manage diagnostics asyncronously
+manage diagnostics asyncronously with linter
--- a/.changeset/sweet-pianos-sniff.md
+++ b/.changeset/sweet-pianos-sniff.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/repl': patch
+---
+
+manage diagnostics asyncronously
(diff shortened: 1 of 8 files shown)


PuruVJ:
Strongly agree that this needs to be in neocodemirror. Will make a related PR

PR #517, opened 2023-07-04
feat(repl): allow dark mode for svelte-json-tree

tanhauhau:
Screenshot 2023-07-05 at 12 05 09 AM

### DIFF
--- a/packages/repl/src/lib/Output/console/ConsoleLine.svelte
+++ b/packages/repl/src/lib/Output/console/ConsoleLine.svelte
@@ -92,24 +92,6 @@
 		font-family: var(--sk-font-mono);
 	}
 
-	@media screen and (prefers-color-scheme: dark) {
-		.log {
-			--json-tree-property-color: #72a2d3;
-			--json-tree-string-color: #6cd1c7;
-			--json-tree-symbol-color: #6cd1c7;
-			--json-tree-boolean-color: #9681f7;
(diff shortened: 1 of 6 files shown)


benmccann (approved):

packages/repl/src/lib/Output/console/ConsoleLine.svelte
```diff
@@ -92,6 +92,24 @@
 		font-family: var(--sk-font-mono);
 	}
 
+	@media screen and (prefers-color-scheme: dark) {
```
PuruVJ, line 95: This wouldn't work well, because of the theme switcher having different theme than system. 

`Repl` component recieves a theme property(light or dark). I believe it can be passed through to ConsoleLine(or added to the context), applied to root element of this component with `class:dark={theme === 'dark'}`

PuruVJ (changes requested):
Thanks a lot for this man!! This was a fairly overlooked thing until now, glad for this change.

I left one comment, which might require a bit of prop drilling. if you could do that it could be great! Otherwise feel free to lemme know, I'll implement it.

packages/repl/src/lib/Output/console/ConsoleLine.svelte
tanhauhau: ok, i've updated it.

PuruVJ (approved):

PR #526, opened 2023-07-10
fix: clarify error message

gtm-nayan:
Without throwing, the error is only logged to the console and the banner shows another confusing error.


packages/repl/src/lib/workers/bundler/index.js
```diff
@@ -286,8 +286,7 @@ async function get_bundle(uid, mode, cache, local_files_lookup) {
 				if (importer && local_files_lookup.has(importer)) {
 					// relative import in a REPL file
 					// should've matched above otherwise importee doesn't exist
-					console.error(`Cannot find file "${importee}" imported by "${importer}" in the REPL`);
-					return;
+					throw new Error(`Cannot find file "${importee}" imported by "${importer}" in the REPL`);
```
PuruVJ, line 289: I remember I did it because the compiler was hanging fully on any error and refusing to re-run on typing. lemme check if this works

benmccann:
sorry this went unreviewed. I'm afraid I'm going to have to close it as the repl code now lives here: https://github.com/sveltejs/svelte.dev/tree/main/packages/repl

PR #528, opened 2023-07-11
feat(hn): update deps, edge based

PuruVJ:
- Updates deps of hn.svelte.dev
- Switches to using edge runtime for faster performance and less impact on environment

### DIFF
--- a/sites/hn.svelte.dev/package.json
+++ b/sites/hn.svelte.dev/package.json
@@ -5,13 +5,17 @@
 	"scripts": {
 		"dev": "vite dev",
 		"build": "vite build",
-		"preview": "vite preview"
+		"preview": "vite preview",
+		"lint": "prettier --ignore-path .gitignore --check --plugin-search-dir=. .",
+		"format": "prettier --ignore-path .gitignore --write --plugin-search-dir=. ."
 	},
 	"devDependencies": {
 		"@sveltejs/adapter-vercel": "^3.0.2",
 		"@sveltejs/kit": "^1.22.2",
 		"browserslist": "^4.21.9",
 		"lightningcss": "^1.21.5",
+		"prettier": "^2.7.1",
+		"prettier-plugin-svelte": "^2.7.0",
 		"svelte": "^4.0.5",
 		"vite": "^4.4.3"
 	},
--- a/sites/hn.svelte.dev/package.json
+++ b/sites/hn.svelte.dev/package.json
@@ -10,8 +10,8 @@
 	"devDependencies": {
 		"@sveltejs/adapter-vercel": "^3.0.2",
 		"@sveltejs/kit": "^1.22.2",
-		"prettier": "^3.0.0",
-		"prettier-plugin-svelte": "^2.10.1",
+		"browserslist": "^4.21.9",
+		"lightningcss": "^1.21.5",
 		"svelte": "^4.0.5",
 		"vite": "^4.4.3"
 	},
--- a/sites/hn.svelte.dev/package.json
+++ b/sites/hn.svelte.dev/package.json
@@ -12,13 +12,11 @@
 	},
 	"devDependencies": {
 		"@sveltejs/adapter-vercel": "^3.0.2",
-		"@sveltejs/kit": "^1.22.2",
-		"browserslist": "^4.21.9",
-		"lightningcss": "^1.21.5",
+		"@sveltejs/kit": "^1.22.3",
 		"prettier": "^2.7.1",
-		"prettier-plugin-svelte": "^2.7.0",
+		"prettier-plugin-svelte": "^2.10.1",
 		"svelte": "^4.0.5",
-		"vite": "^4.4.3"
+		"vite": "^4.4.4"
 	},
 	"type": "module"
 }
--- a/sites/hn.svelte.dev/package.json
+++ b/sites/hn.svelte.dev/package.json
@@ -5,18 +5,15 @@
 	"scripts": {
 		"dev": "vite dev",
 		"build": "vite build",
-		"preview": "vite preview",
-		"lint": "prettier --ignore-path .gitignore --check --plugin-search-dir=. .",
-		"format": "prettier --ignore-path .gitignore --write --plugin-search-dir=. ."
+		"preview": "vite preview"
 	},
 	"devDependencies": {
-		"@sveltejs/adapter-auto": "^2.0.0",
-		"@sveltejs/kit": "^1.5.0",
-		"@sveltejs/vite-plugin-svelte": "^2.0.2",
-		"prettier": "^2.7.1",
-		"prettier-plugin-svelte": "^2.7.0",
-		"svelte": "^3.55.0",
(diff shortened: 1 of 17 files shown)


sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js
```diff
@@ -1,4 +1,3 @@
-/** @type {import('./$types').PageLoad} */
```
Comment (benmccann) [inline sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js:1]: Why remove the types?

Comment (PuruVJ) [reply sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js:1]: Cuz they're automatically inferred by svelte extension, and gives better JSDoc results than manually typing. Its also lesser number of lines

Comment (benmccann) [reply sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js:1]: Will `svelte-check` automatically infer them as well?

Comment (PuruVJ) [reply sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js:1]: It should, svelte.dev doesn't have typings on these functions and lint step passes there

sites/hn.svelte.dev/package.json
```diff
@@ -5,18 +5,15 @@
 	"scripts": {
 		"dev": "vite dev",
 		"build": "vite build",
-		"preview": "vite preview",
-		"lint": "prettier --ignore-path .gitignore --check --plugin-search-dir=. .",
```
benmccann, line 9: why remove `lint` and `format`? we should probably have `check` as well:
```
"check": "svelte-kit sync && svelte-check --tsconfig ./jsconfig.json"
```

sites/hn.svelte.dev/src/routes/user/[name]/+page.js
```diff
@@ -1,7 +1,5 @@
 export const csr = false;
 
-/** @type {import('./$types').PageLoad} */
 export async function load({ params, fetch }) {
-	const res = await fetch(`https://api.hnpwa.com/v0/user/${params.name}.json`);
-	return res.json();
+	return fetch(`https://api.hnpwa.com/v0/user/${params.name}.json`).then((r) => r.json());
```
Comment (benmccann) [inline sites/hn.svelte.dev/src/routes/user/[name]/+page.js:4]: making it one line makes sense, but why not use `await` still?
```suggestion
	return (await fetch(`https://api.hnpwa.com/v0/user/${params.name}.json`)).json();
```

sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js
Comment (benmccann) [reply sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js:1]: Well we don't run `svelte-check` as far as I can see, so that doesn't prove much :laughing: Plus, I don't think it would error on an `any` type - only a known type that was wrong.

@dummdidumm do you know if `svelte-check` is able to infer these types?

sites/hn.svelte.dev/svelte.config.js
```diff
@@ -1,11 +1,14 @@
...
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
 	kit: {
-		adapter: adapter(),
+		adapter: adapter({ runtime: 'edge' }),
 
 		inlineStyleThreshold: 5000
+	},
+	compilerOptions: {
+		cssHash: ({ hash, css }) => `s-${hash(css)}`
```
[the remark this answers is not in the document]
PuruVJ: It makes css files smaller. Currently the output has uses svelte-*. With this its few characters shorter

sites/hn.svelte.dev/src/routes/user/[name]/+page.js
Comment (PuruVJ) [reply sites/hn.svelte.dev/src/routes/user/[name]/+page.js:4]: I dunno, await has felt weird when surrounded by brackets. .then feels cleaner for this case somehow

sites/hn.svelte.dev/package.json
PuruVJ: Force of habbit from my own projects. Will add back

sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js
Comment (PuruVJ) [reply sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js:1]: Just added svelte check and ran it. Works fine here.

AFAIK, any type is not allowed with noImplicitAny, which is on for this project. So svelte-check os detecting the types automatically methinka

sites/hn.svelte.dev/package.json
PuruVJ: Added back, with check

sites/hn.svelte.dev/src/routes/user/[name]/+page.js
Comment (benmccann) [reply sites/hn.svelte.dev/src/routes/user/[name]/+page.js:4]: What happens if invalid JSON is returned from the API? I think the two ways of writing it would behave differently, but I'm not exactly sure what the behavior of the `.then()` version is

sites/hn.svelte.dev/svelte.config.js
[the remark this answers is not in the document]
benmccann: I feel like we probably shouldn't do this in a demo app as it's a bit non-standard and might confuse users. If we think that's better we should probably change the default in Svelte itself

sites/hn.svelte.dev/vite.config.js
```diff
@@ -1,8 +1,19 @@
 import { sveltekit } from '@sveltejs/kit/vite';
+import browserslist from 'browserslist';
+import { browserslistToTargets } from 'lightningcss';
 
 /** @type {import('vite').UserConfig} */
 const config = {
-	plugins: [sveltekit()]
+	plugins: [sveltekit()],
+	css: {
```
benmccann, line 8: I'm kind of unsure about using lightingcss here actually. I think it's fine on svelte.dev/kit.svelte.dev, but it seems to complicate the setup and probably isn't ideal for a demo app. If we think it's better we should probably push to make it the default in SvelteKit or Vite instead

sites/hn.svelte.dev/src/routes/user/[name]/+page.js
Comment (PuruVJ) [reply sites/hn.svelte.dev/src/routes/user/[name]/+page.js:4]: It's the same. If it errors, then .json() errors out, giving the same error as await version

sites/hn.svelte.dev/svelte.config.js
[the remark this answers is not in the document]
PuruVJ: Sure, I'll change it back! I do thing s- is a better alternative. Even Astro doesn't smit svelte-* classes, it shortens it to s-* to produce smaller CSS files

sites/hn.svelte.dev/vite.config.js
PuruVJ: Sure! Will remove

sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js
Comment (PuruVJ) [reply sites/hn.svelte.dev/src/routes/[list=category]/[page]/+page.js:1]: Ah got it! checkJS needs to be true

benmccann (approved):

PR #533, opened 2023-08-18
Exclude HN site from search-engine indexing

CaptainCodeman:
As mentioned in discord, having so many non-svelte results show up when searching for svelte related content can negatively impact people's ability to find the content they are looking for

### DIFF
--- a/sites/hn.svelte.dev/src/app.html
+++ b/sites/hn.svelte.dev/src/app.html
@@ -5,6 +5,7 @@
 		<meta name="description" content="A Hacker News clone built with SvelteKit" />
 		<link rel="icon" href="%sveltekit.assets%/favicon.png" />
 		<meta name="viewport" content="width=device-width, initial-scale=1" />
+		<meta name="robots" content="noindex,nofollow" />
 		<link rel="manifest" href="/manifest.json" />
 		<script>
 			try {


benmccann:
I probably would have implemented this by adding a `robots.txt` in the `static/` directory. It will save a few bytes of traffic for all our real users and also increase our Lighthouse score, which I believe checks if you have defined a `robots.txt`. Since this is an example site, it's probably worth trying to be a little extra pedantic in following the practices we'd like to promote

I'm curious, what searches is hn.svelte.dev showing up for where we'd like it to be excluded?

CaptainCodeman:
The problem with using robots.txt is that it blocks the search engines from being told to remove entries. For a new site it's correct but you need to allow the spiders access to learn that the entries need to be removed (but it could be added after they are)

The example given was for "svelte fuzzing", which for me also brings up a totally non-svelte related page from the hn site:

Screenshot 2023-08-18 at 10 53 11 AM

Looks like there are about 8,250 pages indexed from it.

benmccann:
> The problem with using robots.txt is that it blocks the search engines from being told to remove entries.

wtf! well that explains why I haven't gotten svelte.dev/tutorial out of the search results yet (in favor of learn.svelte.dev). I just requested to remove it in the google search console, but I suppose I'd need to do the same for bing, etc. so this is probably still the better solution in that case so that the other search engines are handled

PR #532, opened 2023-08-16
feat(hn): toggle comments by details > summary

imcotton:
This drops javascript requirement while toggles between show/hide on comments.

PR #535, opened 2023-08-22
fix(hn): correct CSS path for nested comments toggle

imcotton:
Just realized CSS for recursive component not scoped, need to specify its toggle selector.


PuruVJ:
Thanks for the contribution! 

Does this affect the styles in any way?

imcotton:
It's a regression introduced in #532 where only root level's toggle icon (fold/unfold) works, this PR fixed the issue.

PR #546, opened 2023-12-02
chore(repl): Rollup v4

PuruVJ:
Bumps Rollup/browser to v4, which uses WebAssembly.

Comparing, new version seems to load 400KB gzip more overall, which are added from the new Rollup WASM(1.6Mb on its own)

Decision on whether to update to v4 or stay on v3 for the foreseeable future is pending


benmccann:
Yeah, I don't know how I feel about the increased size. I might hold off for the time being

benmccann:
I'm going to go ahead and close this since the repl code now lives here: https://github.com/sveltejs/svelte.dev/tree/main/packages/repl

Issue #457, opened 2023-04-04
[REPL] Switch to jsDelivr


longnguyen2004:
unpkg has been [intermittent](https://github.com/mjackson/unpkg/issues/343) for ~6 months. jsDelivr is more resilient, (seems to) supports CORS and also hosts files from GitHub, so switching to it would fix #282 as well.

jimaek:
Let us know if the jsDelivr team can help with anything. jsDelivr is under active development and aims at production use-cases with multi-CDN logic https://www.jsdelivr.com/network/infographic

theetrain:
In my related comment, I suggested using Skypack to help with package compatibility: https://github.com/sveltejs/sites/issues/137#issuecomment-1360763974

longnguyen2004:
jsDelivr also has [esm.run](https://esm.run), which does ESM conversion and minification as well.

benmccann:
@longnguyen2004 thanks for the suggestion ([as well as the issues in Svelte itself](https://github.com/sveltejs/svelte/issues/created_by/longnguyen2004) - sorry we have a hard time keeping up with all the issues :smile:)

> Let us know if the jsDelivr team can help with anything. jsDelivr is under active development and aims at production use-cases with multi-CDN logic https://www.jsdelivr.com/network/infographic

@jimaek, just curious, how were you made aware of this issue? Do you have a GitHub alert for jsDelivr always running or something?

jimaek:
Yep I monitor jsDelivr mentions in case someone needs help

PuruVJ:
Hi, I'm experimenting with jsDelivr in REPL(See #551). So far loving it, only issue is that it doesn't resolve barrel imports(https://cdn.jsdelivr.net/npm/radix-svelte/dist/components) while unpkg does(https://unpkg.com/radix-svelte/dist/components). 

That is something that would make transitioning very easy. Is this possible to implement on jsdelivr's end? cc @jimaek

MartinKolarik:
> Hi, I'm experimenting with jsDelivr in REPL(See #551). So far loving it, only issue is that it doesn't resolve barrel imports(https://cdn.jsdelivr.net/npm/radix-svelte/dist/components) while unpkg does(https://unpkg.com/radix-svelte/dist/components). Can be done manually in the REPL by resolving with different methods('.js', '.mjs', '.cjs', './index.js', 'index.cjs', 'index.mjs'), but that would require 7 fetch requests in worst case, which is too much to be doing on user's device.
> 
[...]

PR #551, opened 2023-12-29
feat(repl): esm.run as default CDN

PuruVJ:


benmccann:
if my memory serves me, you'd said this didn't work? (I see I left a comment to that effect here: https://github.com/sveltejs/sites/issues/457#issuecomment-1941758472)

but would be happy to get this into the new REPL if it did work. going to close this PR since that code now lives at https://github.com/sveltejs/sites/tree/master/packages/repl

PuruVJ:
JSdelivr will work, but esm.sh bundles everything, which cuts off svelte/* modules having one common component instance

PR #558, opened 2024-05-20
feat: svelte 5 POC

PuruVJ:
Current REPL's state is a mess. Bunch of `get(store)` in .js files. This simplifies all that to use $state

### DIFF
--- a/packages/repl/src/lib/Repl.svelte
+++ b/packages/repl/src/lib/Repl.svelte
@@ -80,12 +80,6 @@
 
 		injectedCSS = data.css || '';
 
-		// when we set new files we also populate the EDITOR_STATE_MAP
-		// with a new state for each file containing the source as docs
-		// this allows the editor to behave correctly when renaming a tab
-		// after having loaded the files externally
-		repl_state.populate_editor_state();
-
 		onchange?.({ files: repl_state.files });
 	}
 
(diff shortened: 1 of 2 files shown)


packages/repl/src/lib/Bundler.js
```diff
@@ -13,7 +13,7 @@ export default class Bundler {
 		const hash = `${packages_url}:${svelte_url}`;
 
 		if (!workers.has(hash)) {
-			const worker = new Worker();
+			const worker = new Worker(bundler_worker, { type: 'module' });
```
paoloricciuti, line 16: I think module workers still doesn't work on Firefox

PuruVJ: https://caniuse.com/?search=worker%20module

benmccann:
It'd probably be easier to review if the prettier upgrade were a separate PR

packages/repl/src/lib/Bundler.js
paoloricciuti: The first result is only for `Worker` if you scroll to the last result `ES Modules in workers` you can see that Firefox doesn't have support yet

PuruVJ:
It has to be done here, otherwise using snippets break prettier across entire project.

packages/repl/src/lib/Bundler.js
PuruVJ: That's for service worker

benmccann:
Couldn't you upgrade prettier first and then do the Svelte 5 upgrade after that?

benmccann:
thank you!

closing this since we have the omnisite, but happy to get any of these changes there

PR #561, opened 2024-10-22
feat(hn): Svelte 5

wackbyte:
Upgrades all other dependencies as well. TypeScript was added as one because it seems that types aren't generated inside `.svelte-kit` otherwise?


benmccann:
thank you!!

PR #568, opened 2025-11-03
chore: use `$app/state`

sacrosanctic:

PR #571, opened 2025-11-03
chore: use `navigating` object property in condition

teemingc:
Follow up to https://github.com/sveltejs/sites/pull/568
Since `navigating` is always an object, we check the truthiness of the object property instead, which can be `null` when there is no ongoing navigation.


benmccann:
The original PR updated `+error.svelte`, so we should probably update that here as well

benmccann (approved):

PR #575, opened 2026-04-06
feat(hn): swap to Algolia and Firebase APIs

### LINKED ISSUE #574: hn.svelte.dev stops updating
Seems something happened on the backend.

GrygrFlzr:
As per davideast/hnpwa-api#103, the currently used HNPWA API is serving stale lists, so I've swapped it over to a combination of the "official" Firebase API and the Algolia one.

The codebase has been relatively modernized:
- Uses `resolve` for all internal paths
- Uses type-safe routes where available
- Safely parse HN's subset of HTML instead of throwing it to `{@html}`
  - Fortunately, they only implement a very small subset: paragraphs, italics, links (not masked links), and code blocks, so the parsing code is relatively simple.

Acknowledged Tradeoffs:
- Accessing an `/item/{id}` will hit both Firebase and Algolia.
  - Algolia contains the entire children tree, whereas the Firebase one requires us to independently fetch every subitem recursively.
  - Firebase provides the real sorting of items, Algolia does not. We only hit Firebase for the top-level item, so deeper nested comments will retain Algolia sorting.
  - It *should* failover if Firebase is down but Algolia is up by just retaining Algolia's original sort.
- Accessing a list such as `/top/1` produces N+1 fetches
  - Firebase's list provides the list of item IDs, Algolia has no respective equivalent, so we're forced to additionally fetch per item to get the title and metadata
  - We do not need the entire descendant tree at this point, so it still uses the Firebase API to do the N queries.
- Uses load functions instead of remote functions
  - While it would serve as an exemplar for how to utilize remote functions, the API is still shifting, so it is more stable at this point to have the project stick to load functions.

Miscellaneous Changes
- Swapped from deprecated vercel `edge` runtime to `node` runtime
- Enabled Vercel ISR at 60 second duration (except for pre-rendered `/about`)
- Bumped `vite-plugin-svelte` and Vite major
- Added `@types/node`
- Stripped `vite-plugin-svelte` preprocess configuration


sites/hn.svelte.dev/src/routes/user/[name]/+page.js
```diff
@@ -1,5 +1,16 @@
-export const csr = false;
+import { browser } from '$app/environment';
+export const csr = browser;
```
Comment (Rich-Harris) [inline sites/hn.svelte.dev/src/routes/user/[name]/+page.js:2]: what does this do?

Comment (GrygrFlzr) [reply sites/hn.svelte.dev/src/routes/user/[name]/+page.js:2]: Intention was if already navigating on browser, keeps routing enabled, previously it just killed all CSR regardless of the prior navigation. I should double check that actually happens...

Comment (Rich-Harris) [reply sites/hn.svelte.dev/src/routes/user/[name]/+page.js:2]: i _think_ it's equivalent to `export const csr = true`, aka the default, though maybe i've got myself confused

PR #577, opened 2026-04-08
feat(hn): internal route for merging multiple remote fetches

GrygrFlzr:
It turns out every page navigation was reloading the remote endpoint fetches.
Both Algolia and Firebase attach `no-cache` in their headers, although I'm not certain that's the root cause specifically.
Either way, it was enough of a problem that I added `[...route]/api` endpoints so that all such accesses appear as single fetches to the pages, which seems to have avoided the problem of over-fetching.
As a bonus, we get to declare our own cache control response headers.

Minor miscellaneous changes
- Actually divide millisecond epoch by 1000 (oops)
- Type fixes for HNItem
- Use `Promise.allSettled` instead of `Promise.all` where partial fetch failure is okay
- `/[list]/rss` uses the `/[list]/1/api` endpoint instead of duplicating the logic

### DIFF
--- a/sites/hn.svelte.dev/src/app.d.ts
+++ b/sites/hn.svelte.dev/src/app.d.ts
@@ -21,6 +21,7 @@ declare global {
 		kids?: number[];
 		score: number;
 		time: number;
+		text?: string;
 		title: string;
 		type: 'story';
 		url: string;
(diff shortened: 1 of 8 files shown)


teemingc (approved):
Thank you!

